PR #1, opened 2019-01-03
PHOENIX-5063 Create a new repo for the phoenix query server

karanmehta93:
@tdsilva @joshelser Please review.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,553 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.apache.phoenix</groupId>
+    <artifactId>phoenix-queryserver</artifactId>
+    <version>4.15.0-HBase-1.4-SNAPSHOT</version>
+    <packaging>pom</packaging>
+    <name>Phoenix Query Server</name>
+
+    <licenses>
+        <license>
+            <name>The Apache Software License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+            <comments/>
+        </license>
+    </licenses>
+
+    <organization>
+        <name>Apache Software Foundation</name>
+        <url>http://www.apache.org</url>
+    </organization>
+
+    <modules>
+        <module>queryserver</module>
+        <module>queryserver-client</module>
+    </modules>
+
+    <repositories>
+        <repository>
+            <id>apache release</id>
+            <url>https://repository.apache.org/content/repositories/releases/</url>
+        </repository>
+    </repositories>
+
+    <scm>
+        <connection>scm:git:http://git-wip-us.apache.org/repos/asf/phoenix.git</connection>
+        <url>https://git-wip-us.apache.org/repos/asf/phoenix.git</url>
+        <developerConnection>scm:git:https://git-wip-us.apache.org/repos/asf/phoenix.git</developerConnection>
+    </scm>
+
+    <properties>
+        <!-- General Properties -->
+        <top.dir>${project.basedir}</top.dir>
+
+        <!-- Hadoop Versions -->
+        <hbase.version>1.4.0</hbase.version>
+        <hadoop-two.version>2.7.5</hadoop-two.version>
+
+        <!-- Dependency versions -->
+        <commons-cli.version>1.2</commons-cli.version>
+        <hive.version>1.2.1</hive.version>
+        <hadoop.version>2.7.1</hadoop.version>
+        <protobuf-java.version>2.5.0</protobuf-java.version>
+        <sqlline.version>1.2.0</sqlline.version>
+        <guava.version>13.0.1</guava.version>
+        <jline.version>2.11</jline.version>
+        <commons-logging.version>1.2</commons-logging.version>
+        <!-- Do not change jodatime.version until HBASE-15199 is fixed -->
+        <avatica.version>1.12.0</avatica.version>
+        <jettyVersion>8.1.7.v20120910</jettyVersion>
+        <servlet.api.version>3.1.0</servlet.api.version>
+        <!-- Test Dependencies -->
+        <mockito-all.version>1.8.5</mockito-all.version>
+        <junit.version>4.12</junit.version>
+
+        <!-- Plugin versions -->
+        <maven-eclipse-plugin.version>2.9</maven-eclipse-plugin.version>
+        <maven-build-helper-plugin.version>1.9.1</maven-build-helper-plugin.version>
+        <maven-surefire-plugin.version>2.20</maven-surefire-plugin.version>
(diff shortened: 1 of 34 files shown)


pom.xml
```diff
@@ -0,0 +1,553 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.apache.phoenix</groupId>
+    <artifactId>phoenix-queryserver</artifactId>
+    <version>4.15.0-HBase-1.4-SNAPSHOT</version>
```
twdsilva: Since this is now in its own repo we should use 1.0.0

pom.xml
```diff
@@ -0,0 +1,553 @@
...
+
+    <repositories>
+        <repository>
+            <id>apache release</id>
+            <url>https://repository.apache.org/content/repositories/releases/</url>
+        </repository>
+    </repositories>
+
+    <scm>
+        <connection>scm:git:http://git-wip-us.apache.org/repos/asf/phoenix.git</connection>
```
twdsilva: these urls need to be updated

pom.xml
```diff
@@ -0,0 +1,553 @@
...
+    </build>
+
+    <dependencyManagement>
+        <dependencies>
+
+            <!-- Intra-project dependencies -->
+            <dependency>
+                <groupId>org.apache.phoenix</groupId>
+                <artifactId>phoenix-core</artifactId>
+                <version>${project.version}</version>
```
twdsilva: Add a new phoenix.version to pom.xml, IMO we should depend on a released version of phoenix (not snapshot).

pom.xml
karanmehta93: Phoenix-connectors started off in this way, so I followed it. But yes, we can update it. We will need to have 4 branches here as well. 1.0.0-HBase-1.4, 1.0.0-HBase-1.3, 1.0.0-HBase-1.2 and master

FYI: https://github.com/apache/phoenix-connectors/blob/4.x-HBase-1.4/presto-phoenix-shaded/pom.xml

pom.xml
karanmehta93: Yes, I feel that too. We need to come up about guidelines for how much version variation can come up for compatibility purposes.

pom.xml
joshelser: > We will need to have 4 branches here as well

Do we need to have four branches? If we're going to make such a change like this already, I'd love to move away from such a burden.

pom.xml
karanmehta93: @joshelser Thoughts here?

pom.xml
```diff
@@ -0,0 +1,553 @@
...
+        <!-- General Properties -->
+        <top.dir>${project.basedir}</top.dir>
+
+        <!-- Hadoop Versions -->
+        <hbase.version>1.4.0</hbase.version>
+        <hadoop-two.version>2.7.5</hadoop-two.version>
+
+        <!-- Dependency versions -->
+        <commons-cli.version>1.2</commons-cli.version>
+        <hive.version>1.2.1</hive.version>
```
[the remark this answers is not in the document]
karanmehta93: Good catch, I did remove many of unused ones. Today I re-evaluated and removed some more and pushed a new commit.

pom.xml
karanmehta93: Yeah, we may not need that. I think we can put a default version of phoenix and override others at runtime for those branches. I believe such a behavior is possible with maven and builds can be setup accordingly.

joshelser: Yah, as long as we're "reasonable" with Phoenix API, we can hopefully avoid all that headache :)

pom.xml
joshelser: Agree to use a release version of Phoenix (e.g. last)

I think this goes to the thread up above too -- choosing one version of Phoenix to test against. Largely, PQS should work against all versions of Phoenix, regardless of "internals". Our choice of phoenix-core to use is more about the version we will run unit tests against (which probably also means we can move this dependency to `<scope>test</scope>` too).

karanmehta93: I didn't quite get your point here @joshelser Can you elaborate?

joshelser: PQS doesn't have a compile-time dependency against Phoenix -- it's only necessary for a Phoenix version when we are building a tarball/packaging for PQS or when we're actually going to run PQS.

I'm thinking that we can make our lives easier by just saying "PQS unit tests run against this version of Phoenix" and running it against different versions is an exercise left to the user.

pom.xml
joshelser: Fine by me.

joshelser:
A couple higher-level thoughts:

* How do I use this? If I have Phoenix 4.14.1 installed, how do I "install PQS"?
* Should PQS overlay the Phoenix installation or create its installation dir?
* This needs a README.md to capture answers on the above.

pom.xml
karanmehta93: Here's what I am saying.

This project will be versioned from `1.0.0-SNAPSHOT`.
The project will have a phoenix.version property which will define the versions of phoenix it will use (phoenix-core)
This value will be defaulted to `4.14.1-HBase-1.4` (Doing so will cause compilation failure since we have a patch in 4.15 (PHOENIX-4755) here, so we need to discuss this)
The compilation can be carried out with `mvn clean install -DskipTests -Dphoenix.version=4.15.0-HBase-1.4-SNAPSHOT` or any other version as deemed appropriate.

pom.xml
karanmehta93: Yes, that makes sense. Refer to my comment for the further plan.

karanmehta93:
> How do I use this? If I have Phoenix 4.14.1 installed, how do I "install PQS"?

We internally use the assembly plugin to create the tar ball, we can have something similar here if we want.

> Should PQS overlay the Phoenix installation or create its installation dir?

Can you elaborate here?

> This needs a README.md to capture answers on the above.

Sure, will add. It would include details about compilation and running for now.

joshelser:
> >  Should PQS overlay the Phoenix installation or create its installation dir?
> Can you elaborate here?

Essentially, is the goal to maintain a `PHOENIX_HOME` that looks exactly like it does today, just built from two different tarballs?

e.g. phoenix-queryserver.tgz would contain files like `bin/queryserver.py`, `phoenix-$VERSION-queryserver.jar`, and `phoenix-$VERSION-thin-client.jar`

karanmehta93:
Yes, I believe that we can configure the plugin appropriately. 
Good catch there, I need to get the relevant files from `bin` directory over here to make that happen. Will add a commit here to address that.

karanmehta93:
I pushed a commit now.
This will cause the compilation failure of the project since it points to the released version `4.14.1-HBase-1.4`, which doesn't have the changes for `PHOENIX-4755`. The project can be successfully compiled with `mvn clean install -DskipTests -Dphoenix.version=4.15.0-HBase-1.4-SNAPSHOT`.

How do you feel we should address that issue? @joshelser @twdsilva

joshelser:
IMO, depend against 4.15.0-SNAPSHOT for now, eventually moving it to 4.15.0 when that's "official".

twdsilva:
+1 on depending on 4.15.0-SNAPSHOT.

twdsilva:
@karanmehta93 We should also just have a single branch master (instead of the current branch 4.x-HBase-1.4)

PR #4, opened 2019-03-31
PHOENIX-5221 Phoenix Kerberos Integration tests failure on Redhat Linux

m2je:
Due to this bug https://bugzilla.redhat.com/show_bug.cgi?id=668830
We need to use `localhost.localdomain` when running these tests on Jenkins (Centos)
but for Mac OS it should be `localhost` to pass.
The reason is kerberos principals in this tests are looked up from /etc/hosts
and 127.0.0.1 is resolved to `localhost.localdomain` rather than `localhost` on Redhat
KDC sees `localhost` != `localhost.localdomain` and as the result test fails with authentication error.
It's also important to note these principals are shared between HDFs and HBase in this mini HBase cluster.
Some more reading https://access.redhat.com/solutions/57330

### DIFF
--- a/queryserver/src/it/java/org/apache/phoenix/end2end/AbstractKerberisedTest.java
+++ b/queryserver/src/it/java/org/apache/phoenix/end2end/AbstractKerberisedTest.java
@@ -1,0 +1,283 @@
+package org.apache.phoenix.end2end;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
+import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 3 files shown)


karanmehta93:
@m2je Thanks for the patch and apologies about the delay in review. 
I see a lot of common code in the test files, can you also refactor the common code to a base test class instead?

m2je:
@karanmehta93 Done with the comments. Please review again.

m2je:
@karanmehta93 Done.
Thanks

PR #5, opened 2019-04-23
PHOENIX-5255: Create Orchestrator for QueryServerCanaryTool in phoenix-queryserver project

swaroopak:
This PR aims at implementing an orchestrator for PQS canary, such that it executes a canary tool which runs UPSERT and READ in a Phoenix table at every configurable time interval via PQS.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,4 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+
+-->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
@@ -30,6 +50,7 @@
         <module>load-balancer</module>
         <module>assembly</module>
         <module>phoenix-client</module>
+        <module>queryserver-orchestrator</module>
     </modules>
 
     <repositories>
@@ -53,7 +74,7 @@
         <hbase.version>1.4.0</hbase.version>
         <hadoop-two.version>2.7.5</hadoop-two.version>
         <curator.version>2.12.0</curator.version>
-        <phoenix.version>4.15.0-HBase-1.4-SNAPSHOT</phoenix.version>
+        <phoenix.version>4.15.0-HBase-1.3-SNAPSHOT</phoenix.version>
 
         <!-- Dependency versions -->
         <protobuf-java.version>2.5.0</protobuf-java.version>
@@ -233,7 +254,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-shade-plugin</artifactId>
-                    <version>2.4.3</version>
+                    <version>3.2.0</version>
                 </plugin>
                 <plugin>
                     <!-- Allows us to get the apache-ds bundle artifacts -->
@@ -508,6 +529,16 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.curator</groupId>
+                <artifactId>curator-recipes</artifactId>
+                <version>${curator.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.curator</groupId>
+                <artifactId>curator-framework</artifactId>
+                <version>${curator.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.curator</groupId>
                 <artifactId>curator-test</artifactId>
                 <version>${curator.version}</version>
             </dependency>
@@ -570,6 +601,16 @@
                 <groupId>javax.servlet</groupId>
                 <artifactId>javax.servlet-api</artifactId>
                 <version>${servlet.api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-api</artifactId>
+                <version>1.8.0-alpha2</version>
+            </dependency>
+            <dependency>
+                <groupId>net.sourceforge.argparse4j</groupId>
+                <artifactId>argparse4j</artifactId>
+                <version>0.8.1</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/queryserver-orchestrator/pom.xml
+++ b/queryserver-orchestrator/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>phoenix-queryserver</artifactId>
+        <groupId>org.apache.phoenix</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>queryserver-orchestrator</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.phoenix</groupId>
+            <artifactId>queryserver-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.calcite.avatica</groupId>
+            <artifactId>avatica-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>net.sourceforge.argparse4j</groupId>
+            <artifactId>argparse4j</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-framework</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-recipes</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.phoenix</groupId>
+            <artifactId>phoenix-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
+++ b/queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
@@ -1,0 +1,141 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.queryserver.orchestrator;
+
+import net.sourceforge.argparse4j.ArgumentParsers;
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.Namespace;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.BoundedExponentialBackoffRetry;
+
+import org.apache.curator.utils.CloseableUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/*
+    This is a main class to which initiates the relevant clients
+    to run orchestrator for queryserver canary
+ */
+
+public class QueryServerCanaryOrchestrator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(QueryServerCanaryOrchestrator.class);
+
+    private static String ZK_PATH = "/pqs/leader";
+
+    private static Namespace parseArgs(String[] args) {
+
+        ArgumentParser parser = ArgumentParsers.newFor("PQS Canary Orchestrator").build()
+                .description("PQS Canary Orchestrator");
+
+        parser.addArgument("--zkurl", "-zk").type(String.class).help("URL for Zookeeper");
+
+        parser.addArgument("--zkpath", "-zkp").type(String.class).nargs("?").setDefault(ZK_PATH)
+                .help("ZKNode path default: " + ZK_PATH);
+
+        parser.addArgument("--hostname", "-hn").type(String.class).nargs("?").setDefault("localhost").help("Hostname on "
+                + "which PQS is running.");
+
+        parser.addArgument("--port", "-p").type(String.class).nargs("?").setDefault("8765").help("Port on which PQS " +
+                "is running.");
+
+        parser.addArgument("--constring", "-cs").type(String.class).nargs("?").help("Pass an " +
+                "explicit " + "connection String to connect to PQS. default: null");
+
+        parser.addArgument("--timeout", "-t").type(String.class).nargs("?").setDefault("120")
+                .help("Maximum time for which the app should run before returning error. default: 120 sec");
+
+        parser.addArgument("--testtable", "-tt").type(String.class).nargs("?").setDefault
+                ("PQSTEST").help("Custom name for the test table. default: PQSTEST");
+
+        parser.addArgument("--testschema", "-ts").type(String.class).nargs("?").setDefault
+                ("TEST").help("Custom name for the test table. default: TEST");
+
+        parser.addArgument("--logsinkclass", "-lsc").type(String.class).nargs("?").setDefault
+                ("org.apache.phoenix.tool.PhoenixCanaryTool$StdOutSink")
+                .help("Path to a Custom implementation for log sink class. default: stdout");
+
+        parser.addArgument("--interval", "-in").type(String.class).nargs("?").setDefault("900")
+                .help("Time interval between 2 consecutive test suite runs");
+
+
+        Namespace res = null;
+        try {
+            res = parser.parseArgs(args);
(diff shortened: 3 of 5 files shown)


swaroopak:
@karanmehta93 FYI

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/LoggingCanarySink.java
```diff
@@ -0,0 +1,75 @@
+package org.apache.phoenix.queryserver.orchestrator;
```
karanmehta93: nit: license missing in couple files

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>queryserver-orchestrator</artifactId>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
```
karanmehta93: I think it should inherit this plugin from the parent pom and it should be needed to explicitly specify it here.

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-release-plugin</artifactId>
```
karanmehta93: We don't usually specify release plugin version in the pom AFAIK. @joshelser Any ideas on this one?

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-recipes</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.phoenix</groupId>
+            <artifactId>phoenix-core</artifactId>
+        </dependency>
+    </dependencies>
+    <dependencyManagement>
```
karanmehta93: Do we need this section? `dependencyManagement` sections are usually in parent pom files

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/LoggingCanarySink.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.phoenix.tool.PhoenixCanaryTool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+public class LoggingCanarySink implements PhoenixCanaryTool.Sink {
```
karanmehta93: nit: please add descriptive comments for all relevant classes and methods

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.curator.retry.BoundedExponentialBackoffRetry;
+
+import org.apache.curator.utils.CloseableUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class QueryServerCanaryOrchestrator {
```
karanmehta93: nit: can you switch the indentation to spaces instead of tabs as per the overall code style?

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,117 @@
...
+
+	private static Namespace parseArgs(String[] args) {
+
+		ArgumentParser parser = ArgumentParsers.newFor("PQS Canary Orchestrator").build()
+				.description("PQS Canary Orchestrator");
+
+		parser.addArgument("--zkpath", "-zkp").type(String.class).nargs("?").setDefault(PATH)
+				.help("ZKNode path default: " + PATH);
+
+		parser.addArgument("--hostname", "-hn").type(String.class).nargs("?").help("Hostname on "
```
karanmehta93: Will be good to add defaults here for hostname and port.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,117 @@
...
+			Runtime.getRuntime().addShutdownHook(new Thread() {
+				public void run() {
+					logger.info("Shutdown Hook for PQS Canary Orchestrator running...");
+					CloseableUtils.closeQuietly(executorClient);
+					CloseableUtils.closeQuietly(client);
+					logger.info("Closed Curator Client");
+				}
+			});
+
+			while (true) {
```
karanmehta93: This causes 100% CPU utilization and should not be used. Ideally this process should spawn off a single new thread and call the join method here. https://stackoverflow.com/questions/14763434/how-to-avoid-100-cpu-utilization-without-removing-whiletrue

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,117 @@
...
+			CuratorFramework client = CuratorFrameworkFactory.newClient(params.get("zkurl"), new
+					BoundedExponentialBackoffRetry(1000, 60000, 50000));
+
+
+			TestExecutorClient executorClient = new TestExecutorClient(client, PATH, params, tool);
+			client.start();
+			executorClient.start();
+
+			// A shutdown hook to cleanly terminate the application
+			Runtime.getRuntime().addShutdownHook(new Thread() {
```
karanmehta93: Do we need to mark this thread as daemon?

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
```diff
@@ -0,0 +1,122 @@
...
+
+            TimeUnit.SECONDS.sleep(Integer.parseInt(params.get("interval")));
+        }
+    }
+
+    @Override
+    public void stateChanged(CuratorFramework client, ConnectionState newState) {
+        logger.info("ZK Connection State Changed to [{}]", newState.name());
+        switch(newState) {
+            case CONNECTED:
```
karanmehta93: Does the `CONNECTED` state mean that the host has leadership?

pom.xml
```diff
@@ -30,6 +30,7 @@
         <module>load-balancer</module>
         <module>assembly</module>
         <module>phoenix-client</module>
+        <module>queryserver-orchestrator</module>
```
joshelser: If this is for phoenix-queryserver, why is this in the main phoenix repo. Could be that I just don't understand what the orchestrator is used for, but it seems like this should be in the new repo: https://github.com/apache/phoenix-queryserver

queryserver-orchestrator/pom.xml
joshelser: The parent pom should contain pluginManagement/dependencyManagement sections which define the versions of plugins/dependencies used in the modules. Be aware that the ASF parent pom also defines plugins, so you often do not need to set one by hand.

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+                <version>2.5.3</version>
+            </plugin>
+        </plugins>
+    </build>
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.8.0-alpha2</version>
+            <scope>compile</scope>
```
joshelser: Drop scope.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,117 @@
...
+					CloseableUtils.closeQuietly(client);
+					logger.info("Closed Curator Client");
+				}
+			});
+
+			while (true) {
+			}
+
+		} catch (Exception e) {
+			logger.error("Error in PQS Canary Orchestrator. " + e);
```
joshelser: rethrow the exception.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
```diff
@@ -0,0 +1,122 @@
...
+                logger.info("The host became the leader at " +System.currentTimeMillis());
+                break;
+            case LOST:
+                logger.info("The host lost leadership at "+ System.currentTimeMillis());
+                throw new CancelLeadershipException();
```
joshelser: This is likely going to be thrown in a separate thread. You need to interrupt the thread which is running your tool.

karanmehta93:
@swaroopak Can you add brief description and motivation behind this in this PR, which will help others to review? Thanks

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/LoggingCanarySink.java
swaroopak: done

queryserver-orchestrator/pom.xml
swaroopak: done

queryserver-orchestrator/pom.xml
swaroopak: done

queryserver-orchestrator/pom.xml
swaroopak: yet to be addressed.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/LoggingCanarySink.java
swaroopak: yet to be addressed.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: done

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: done

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: done

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: Yes, done.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
swaroopak: No, it means it was able to connect under the zkpath. Updated the comment.

queryserver-orchestrator/pom.xml
swaroopak: done

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: done

pom.xml
swaroopak: Yes, it is under new repo phoenix-queryserver.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
swaroopak: yet to be addressed.

swaroopak:
@karanmehta93 @joshelser I have updated PR by addressing (1/2) comments. 
An upcoming commit should include the following:
1. pom.xml update
2. Unit test and IT for the orchestrator
3. Descriptive comments on the classes
4. And https://github.com/apache/phoenix-queryserver/pull/5/files/2e3978519f19958f63391fd11fd990c5c2b3b00b#diff-99377e07a8f1b02ab56286db1c29f05cR97

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,136 @@
...
+            testExecutorThread.start();
+
+            Thread cleanupThread = new Thread(() -> {
+                LOGGER.info("Shutdown Hook for PQS Canary Orchestrator running...");
+                CloseableUtils.closeQuietly(testExecutorClient);
+                CloseableUtils.closeQuietly(curatorClient);
+                LOGGER.info("Closed Curator Client");
+            });
+            cleanupThread.setDaemon(true);
+            cleanupThread.start();
```
swaroopak: Test the sequence.

pom.xml
```diff
@@ -53,7 +54,7 @@
         <hbase.version>1.4.0</hbase.version>
         <hadoop-two.version>2.7.5</hadoop-two.version>
         <curator.version>2.12.0</curator.version>
-        <phoenix.version>4.15.0-HBase-1.4-SNAPSHOT</phoenix.version>
+        <phoenix.version>4.15.0-HBase-1.3-SNAPSHOT</phoenix.version>
```
karanmehta93: This should stay to `4.15.0-HBase-1.4-SNAPSHOT` version. Please don't update it.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,144 @@
...
+                CloseableUtils.closeQuietly(curatorClient);
+                LOGGER.info("Closed Curator Client");
+            });
+            cleanupThread.setDaemon(true);
+            try {
+                testExecutorThread.join();
+            } catch (InterruptedException ex) {
+                LOGGER.error("The Main thread was interrupted", ex);
+            } finally {
+                cleanupThread.start();
```
karanmehta93: The main thread will exit here and will kill the cleanupThread too, since its a daemon thread. I think it's better to make this cleanup synchronous and spawning a new thread is not providing any benefit here.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+
+            Map<String, String> params = getArgs(cArgs);
+            PATH = params.get("zkpath");
+
+            CuratorFramework curatorClient = CuratorFrameworkFactory.newClient(params.get("zkurl"), new
+                    BoundedExponentialBackoffRetry(1000, 60000, 50000));
+            curatorClient.start();
+
+            TestExecutorClient testExecutorClient = new TestExecutorClient(curatorClient, PATH, params, tool);
```
swaroopak: Change the TestExecutorClient to runnable

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+
+public class QueryServerCanaryOrchestrator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(QueryServerCanaryOrchestrator.class);
+
+    private static String PATH = "/pqs/leader";
+
+    private static Namespace parseArgs(String[] args) {
+
+        ArgumentParser parser = ArgumentParsers.newFor("PQS Canary Orchestrator").build()
```
karanmehta93: If possible, can you use any other standard library to parse arguments?
 Checkout how phoenix client tools currently parse args.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
karanmehta93: It would be good to have `TestExecutorClient` as a thread itself and launch it directly or via ExecutorService with single thread pool.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: https://github.com/apache/phoenix/blob/master/phoenix-core/pom.xml#L395

Looks like we are using the same library.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: I could make it runnable and create a new thread to run the client. It is already extending a class, it won't be able to extend another, but implement the Runnable interface.

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <parent>
+        <artifactId>phoenix-queryserver</artifactId>
+        <groupId>org.apache.phoenix</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>queryserver-orchestrator</artifactId>
+
+    <dependencies>
```
karanmehta93: Is the `queryserver-client` dependency missing here?

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: Since the dependent thread is a curator client which depends on the framework to call takeLeadership() method. This process is oblivious to the thread that it might depend on, hence the join on testExecutorClient will see the thread execution is completed and let main thread exit. 

Any suggestion on how do we handle this here?

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
joshelser: Please do not concatenate the exception, but use the logger method that accepts an exception. e.g. `logger.error(message, exception)`

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
joshelser: You want to implement some logic in `TestExecutorClient` that this caller can block on.

For `Thread`, this would be doing something like calling `join()` on a thread you started which you want to wait to complete. Another way to do this would be to make a boolean method `isDone()`on `TestExecutorClient` which this class can call periodically, e.g.
```
while (!client.isDone()) {
  Thread.sleep(1, TimeUnit.SECOND);
}
```

queryserver-orchestrator/pom.xml
joshelser: Doesn't look like this was done..

queryserver-orchestrator/pom.xml
joshelser: Does not look done.

pom.xml
joshelser: Ok, issue is not with you. I'm just confused why there is a `phoenix-client` module for the queryserver repository. Sorry about that.

queryserver-orchestrator/pom.xml
swaroopak: added

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
karanmehta93: Looks like this isn't addressed yet. The code may still continue running in split brain scenarios. I would suggest splitting the PR into two parts, and submit the leader election part later. How do you want to proceed?

queryserver-orchestrator/pom.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
karanmehta93: nit: add license

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/QueryServerCanaryOrchestrator.java
swaroopak: Thank you, @joshelser. 
join didnt help this scenario so used other approach.

queryserver-orchestrator/src/main/java/org/apache/phoenix/queryserver/orchestrator/TestExecutorClient.java
swaroopak: Good point, @karanmehta93 I ll add a TODO and point to new jira to handle this scenario. This needs extra attention.

PR #7, opened 2019-07-17
PHOENIX-5394 Fix the build so it actually runs

joshelser:


karanmehta93:
@joshelser, Why did we remove phoenix-client from pom? I think I added it because I wanted it to be bundled up in the final tar bar generated by mvn assembly. If that still works and the scripts in bin folder work too, then I am good.
Overall, patch is LGTM.

joshelser:
> Why did we remove phoenix-client from pom?

You defined a new module called phoenix-client which is not how you get the phoenix-client jar from phoenix.git.

But to address what you wanted it to do, it's better to come up with a solution that starts PQS by specifying the Phoenix-client jar you want to use. PQS should be able to work with whatever jar you provide at runtime.

karanmehta93:
Yes, that's right. I used to switch jars during compile time using the maven option `-Dphoenix.version`.  I wonder what will be implications of compiling vs run time jars are different. In theory, yes, PQS should not be affected by it.

joshelser:
> I wonder what will be implications of compiling vs run time jars are different. In theory, yes, PQS should not be affected by it.

Yeah, something to validate. Need to come up with some kind of infrastructure to test this, I reckon.

Thanks for looking, let me push this and get back to it.

PR #13, opened 2019-12-19
PHOENIX-5641 Decouple phoenix-queryserver from phoenix-core

stoty:


queryserver/src/main/java/org/apache/phoenix/util/InstanceResolver.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Resolves object instances registered using the JDK 6+ {@link java.util.ServiceLoader}.
+ *
+ * 
+ * @since 2.0
+ */
+public class InstanceResolver {
```
joshelser: This class has been a pain for me in the past. If we don't need it, we might be able to simplify things :). Fine for this PR, just sharing some background :)

stoty:
Thanks Josh. 
Not yet, I wanted to wait until this side is merged. I expect to do that tomorrow.

joshelser:
Sounds great! Thanks for being thorough in this cleanup.

PR #19, opened 2020-03-09
PHOENIX-5761 sqlline-thin kerberos logic too aggressive

stoty:


stoty:
The problem is the BASIC the default authentication method in Avatica, and we are are overriding it based on hadoop/hbase configuration we find on the host.
We could parse the URL, and not start the kerberos login if 
a.) explicit non-SPNEGO is auth is specified
b.) explicit principal is specified (for keytab auth)
c.) explicit auth user is set (not used for SPENGO)

This would probably cover most bases.

joshelser:
> The problem is the BASIC the default authentication method in Avatica, and we are are overriding it based on hadoop/hbase configuration we find on the host.

Oh, I like that (parsing the JDBC url). That should be much more "consumable" by us, than trying to parse command line arguments :)

stoty:
On a second thought, I am not actually sure if this code is useful.
AFAICT it won't log us in if we are not yet logged in, and the Hadoop structures that are initialized are not used the Avatica client. 
What would we lose by just dropping the kerberos logic ?

stoty:
BTW, I've merged this in the meantime, thanks for the review

joshelser:
> AFAICT it won't log us in if we are not yet logged in, and the Hadoop structures that are initialized are not used the Avatica client.
> What would we lose by just dropping the kerberos logic ?

My original goal was such that the following would work naturally:
```
$ kinit jelser
Enter password for jelser@REALM:
$ phoenix-sqlline-thin http://localhost:8765;AUTHENTICATION=SPNEGO
```
The more I think about this, I'm wondering if this is necessary. There might be something in the thin shim layer around the Avatica jdbc client we have in Phoenix? A good thread to pull on... I don't remember anymore.

I remember writing this SqllineWrapper class out of necessity to make it work "similarly" to sqlline.py, but, with PQS behind http proxy's being more and more popular with folks, I think we might just want to rethink it entirely :)

PR #20, opened 2020-03-12
PHOENIX-5772 Streamline the kerberos logic in thin client java code

stoty:
move hbase config parameter processing to python
remove hadoop dependency from queryserver-client
use shaded avatica client everywhere
update sqlline to 1.9
don't shade sqlline + deps into thin client
remove unnecessary interdependencies between modules


bin/thin_client_jaas.conf
```diff
@@ -0,0 +1,5 @@
+ThinClient {
+  com.sun.security.auth.module.Krb5LoginModule required
```
joshelser: I _think_ this will break on IBM java. They do something annoying like have Krb5LoginModule in a different package, I think `com.ibm.security.auth.module.Krb5LoginModule`.

Also annoying, the semantics on what you options you may/must provide in the configuration block are different between vendors.

bin/thin_client_jaas.conf
```diff
@@ -0,0 +1,5 @@
+ThinClient {
```
joshelser: Maybe make this PhoenixThinClient? Disambiguate from other products who have JAAS config names.

e.g. ZooKeeper uses a generic name like "Client" or something equally annoying and likely to conflict with other systems :)

bin/thin_client_jaas.conf
joshelser: I think these options are OK...

https://www.ibm.com/support/knowledgecenter/SSYKE2_7.1.0/com.ibm.java.security.api.71.doc/jgss/com/ibm/security/auth/module/Krb5LoginModule.html seems to indicate that it's actually `useKeytab` on IBM.

Probably easiest to just acknowledge that we only expect this to work on openjdk/oraclejdk (and maybe azul?). All of those have the exact same JAAS config semantics for krb5, AFAIK.

queryserver-client/src/main/java/org/apache/phoenix/queryserver/client/SqllineWrapper.java
```diff
@@ -28,71 +33,55 @@
...
-        System.err.println("Kerberos login failed using ticket cache. Did you kinit?");
+  static Subject login() throws LoginException {
+    Subject subject = new Subject();
+
+    LoginContext lc;
+    lc = new LoginContext("ThinClient", subject, new CallbackHandler() {
+      @Override
+      public void handle(Callback[] callbacks)
+          throws IOException, UnsupportedCallbackException {
+        throw new UnsupportedCallbackException(callbacks[0], "Only ticket cache is supported");
```
joshelser: `-Djavax.security.auth.useSubjectCredsOnly=false` should also prevent JAAS from trying any other ticket acquisition method if the ticket cache login fails.

Even if setting that system property works, having an clear Unsupported exception here is good!

queryserver-client/src/main/java/org/apache/phoenix/queryserver/client/SqllineWrapper.java
```diff
@@ -28,71 +33,55 @@
...
-    final Configuration conf = new Configuration(false);
-    conf.addResource("hbase-site.xml");
+    String url = getUrl(args);
 
-    // Check if the server config says SPNEGO auth is actually disabled.
-    final boolean disableSpnego = conf.getBoolean(QUERY_SERVER_SPNEGO_AUTH_DISABLED_ATTRIB,
-        DEFAULT_QUERY_SERVER_SPNEGO_AUTH_DISABLED);
-    if (disableSpnego) {
-      SqlLine.main(args);
+    if(url.contains("authentication=SPNEGO;") && !url.contains("principal=")) {
```
joshelser: Maybe just `"authentication=SPNEGO"`? This would fail if authentication is the last element (maybe this is the problem you were having above?

stoty: That's it!

bin/thin_client_jaas.conf
stoty: I did, but since it's internal only, I don't think it matters much.

stoty: I mean I've moved this to java code.

bin/thin_client_jaas.conf
stoty: You are right.
I cobbled together a solution without external config file, and with IBM JRE support based on the hadoop minicluster config. It mostly works with IBM JRE now.
Getting the IBM JRE actually talk to a HTTPS server was also fun. (Only old-timers remember the export-only limited cipher brouhaha, that IBM still has)

bin/sqlline-thin.py
```diff
@@ -164,15 +181,34 @@ def get_serialization():
...
 if args.truststore:
     jdbc_url += ';truststore=' + args.truststore
 if args.truststore_password:
     jdbc_url += ';truststore_password=' + args.truststore_password
 
+
+# Add SPENGO auth if this cluster uses it, and there are no conflicting HBase parameters
+if (get_hbase_authentication() == 'kerberos' and get_spnego_auth_disabled() == 'false'
+   and 'authentication=' not in jdbc_url and 'avatica_user=' not in jdbc_url):
+    # For some bizarre reason Avatica doesn't seem to accept authentication as the last parameter
```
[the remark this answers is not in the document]
stoty: It was super simple, as you found it at first look :)

bin/sqlline-thin.py
```diff
@@ -164,15 +181,34 @@ def get_serialization():
     jdbc_url += ';avatica_user=' + args.auth_user
 if args.auth_password:
     jdbc_url += ';avatica_password=' + args.auth_password
+if args.principal:
```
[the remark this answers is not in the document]
stoty: In theory we should. But then we should also parse the passed-in URL, and do something with the conflicts between that and the cli options, and I did not want to go that deep.

bin/sqlline-thin.py
```diff
@@ -92,25 +100,34 @@ def get_serialization():
...
+        return default
 
     stdout = stdout.strip()
     if stdout == 'null':
-        return default_serialization
+        return default
     return stdout
 
+def get_serialization():
+    return get_hbase_param(serialization_key, default_serialization)
```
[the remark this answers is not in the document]
stoty: We only call these methods once, so there is no point.
Pulling a full effective config, and then parsing all options from there would save us noticeable startup time, but I could not find a way to do that.

stoty:
Some general notes:

We don't repackage sqlline, we just use the pre-packaged uberjar provided by them.

You did not spell out exactly in your summary, but the major win here is using the pre-shaded avatcia jar instead of shading ourselves, and omitting hbase, hadoop, and sqlline from the JDBC driver JAR.

PR #27, opened 2020-04-16
PHOENIX-5844 Feature parity for the python client (part 1:authentication rewrite)

stoty:


stoty:
Thanks @joshelser 

Yes, I've tested it manually, both with and without Knox.

I am looking at getting the Python test suite working, but it seems that some refactoring is needed to allow for the auth options.

joshelser:
👍 perfect. Thanks for your due diligence!

PR #29, opened 2020-04-18
PHOENIX-5873 Fix loadbalancer packaging problem for queryserver scripts

infraio:


infraio:
@stoty Any comments?

stoty:
It looks solid, but doesn't address every improvement that I have in mind for this ticket.
Could you make a different JIRA, say "Fix loadbalancer packaging" for it, and rename the patch ?

joshelser:
Looks fine to me -- also, would echo what Istvan already said about fixing the description here to be more accurate after the other changes which have been done.

Can do those on commit if you'd prefer, @infraio

stoty:
I'd like a new JIRA, not just a description, because there are still more problems that I've enumerated in this JIRA, and I'd prefer not to "use up" this JIRA on the loadbalancer fixes.

joshelser:
> I'd like a new JIRA, not just a description, because there are still more problems that I've enumerated in this JIRA

OK, you got it Istvan. Sorry for making you come back here again and requesting what you wanted again.

joshelser:
Looks like this has landed. Good to close, @stoty?

stoty:
Sure.

stoty:
Committed.

PR #30, opened 2020-04-20
PHOENIX-5844 Feature parity for the python client (part2: transaction…

stoty:
… support and test suite)

includes:
PHOENIX-5857 Get the python test suite working
PHOENIX-5858 Add commit/rollback support to the python client


queryserver-it/src/it/java/org/apache/phoenix/end2end/SecureQueryServerPhoenixDBIT.java
```diff
@@ -324,6 +324,18 @@ public static synchronized void stopKdc() throws Exception {
 
     @Test
     public void testBasicReadWrite() throws Exception {
+        File file = new File(".");
+        runShellScript("python", Paths.get(file.getAbsolutePath(), "src", "it", "bin", "test_phoenixdb.py").toString());
```
joshelser, line 328: If it's easier, we could also bind some action to `pre-integration-test` and `post-integration-test` which start a PQS instance. I remember trying to debug failing python tests were a major pain in the rear with the I/O redirection flowing back through the Java maven process.

joshelser:
I guess we don't have a good way to run automated PQS python tests? If you've run them locally and they pass, then you have my +1. I don't have the setup locally to test this out right now.

stoty:
Yes, I did run the test.

Now I made another pass on the tests, and properly integrated the python test suite into the maven test run.

I had actually made some enhancements that helps with development:

- The python output of the maven is tee-d into separate files so that it is more readable
- You can start test server with QueryServerBasicsIT.startLocalPQS
- You can start a HTTPS+kerberos server with SecureQueryServerPhoenixDBIT.startLocalPQS, it will even output a script for you to set up kerberos and the python test suite.

joshelser:
> I had actually made some enhancements that helps with development:

That's wonderful. Thank you!

stoty:

joshelser:
A follow-on maybe. With python 3.7.6, I see the following UT failure:
```
======================================================================
FAIL: test_parse_url (phoenixdb.tests.test_avatica.ParseUrlTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jelser/projects/phoenix-queryserver.git/python/phoenixdb/phoenixdb/tests/test_avatica.py", line 24, in test_parse_url
    self.assertEqual(urlparse.urlparse('http://localhost:2222/'), parse_url('localhost:2222'))
AssertionError: ParseResult(scheme='http', netloc='localhost:2222', path='/',[29 chars]t='') != ParseResult(scheme='localhost', netloc='', path='2222', param[23 chars]t='')
```
Seems like it might be unique to this version of Python?

stoty:
Yes, it seems to be unique.
I couldn't repro it either locally, or with the tox multi-version tests.

I didn't test with 3.7.6, as I have already upgraded to 3.7.7

PR #33, opened 2020-05-15
PHOENIX-5826 Remove guava from queryserver

stoty:


stoty:
The non-trivial parts are:

Copying HostAndPort with minimal changes, to ensure backward compatibility in the ZK string format

The SimpleLRUCache implementation that is used instead of Guava cache.

queryserver/src/main/java/org/apache/phoenix/util/HostAndPort.java
```diff
@@ -0,0 +1,284 @@
+/*
+ * Copyright (C) 2011 The Guava Authors
```
joshelser, line 2: We should ensure that a top-level NOTICE file propagates this copyright notice, as well as the containing jar.

the ASF parent pom already has convention to append entries to the NOTICE file. One example https://github.com/apache/hbase/blob/master/hbase-thrift/src/main/appended-resources/META-INF/NOTICE. The "src/main/appended-resources" is the magic set up by asf.pom.

queryserver/src/main/java/org/apache/phoenix/util/HostAndPort.java
```diff
@@ -0,0 +1,284 @@
...
+ *   <li>[2001:db8::1]:80  - {@link #getHostText()} omits brackets
+ *   <li>2001:db8::1       - Use {@link #requireBracketsForIPv6()} to prohibit this
+ * </ul>
+ *
+ * <p>Note that this is not an exhaustive list, because these methods are only
+ * concerned with brackets, colons, and port numbers.  Full validation of the
+ * host field (if desired) is the caller's responsibility.
+ *
+ * @author Paul Marks
+ * @since 10.0
```
joshelser, line 55: Can we include a comment somewhere in here which states what version of Guava this was copied from?

queryserver/src/main/java/org/apache/phoenix/util/SimpleLRUCache.java
```diff
@@ -0,0 +1,97 @@
...
+            if (currentSize <= triggerSize) {
+                return;
+            }
+            LOG.warn("UGI Cache capacity exceeded, you may want to increase its size");
+            TreeSet<Entry<Object, AtomicLong>> sortedByLRU = new TreeSet<>(
+                    new Comparator<Entry<Object, AtomicLong>>() {
+                        @Override
+                        public int compare(Entry<Object, AtomicLong> o1, 
+                                Entry<Object, AtomicLong> o2) {
+                            if( Long.compare(o2.getValue().get(), o1.getValue().get()) == 0) {
```
joshelser, line 78: Compute this once, and use it for your `if` and `return`

queryserver/src/main/java/org/apache/phoenix/util/HostAndPort.java
stoty: Yes, done.

queryserver/src/main/java/org/apache/phoenix/util/SimpleLRUCache.java
stoty: Ok, done.

stoty:
Making sure that the licensing documentation is in order seems to be a bigger task than the whole Guava removal.

I have split that out into https://issues.apache.org/jira/browse/PHOENIX-5901

stoty:
I forgot to actually push the changes. Better late than never.

stoty:
Thank you @joshelser .
I wasn't comfortable with the comparator returning inconsistent results when the timestamp is equal, so I've changed the implementation a bit.
Could you re-check the last commit ?

joshelser:
> Could you re-check the last commit ?

Looks fine. Thanks for double checking.

PR #35, opened 2020-05-20
PHOENIX-5907 Remove unused part from phoenix_utils.py

infraio:


bin/queryserver.py
```diff
@@ -121,12 +119,11 @@
 #    " -XX:+UnlockCommercialFeatures -XX:+FlightRecorder -XX:FlightRecorderOptions=defaultrecording=true,dumponexit=true" + \
 
 # The command is run through subprocess so environment variables are automatically inherited
-java_cmd = '%(java)s -cp ' + hbase_config_path + os.pathsep + hadoop_config_path + os.pathsep + \
+java_cmd = '%(java)s -cp ' + hbase_config_path + os.pathsep + \
```
stoty, line 122: I don't think that we should remove hadoop_config_path
We need the Hadoop config files for operation, and there is no guarantee that the hbase config dir contains a copy. It seems to be deployment dependent.

See https://hbase.apache.org/book.html#fully_dist

bin/phoenix_utils.py
```diff
@@ -113,60 +110,10 @@ def setPath():
...
+    phoenix_client_jar = find(PHOENIX_CLIENT_JAR_PATTERN, phoenix_class_path)
     if phoenix_client_jar == "":
         phoenix_client_jar = findFileInPathWithoutRecursion(PHOENIX_CLIENT_JAR_PATTERN, os.path.join(current_dir, ".."))
-    if phoenix_client_jar == "":
-        phoenix_client_jar = find(PHOENIX_CLIENT_JAR_PATTERN, phoenix_class_path)
-
-    global phoenix_test_jar_path
-    phoenix_test_jar_path = os.path.join(current_dir, "..", "phoenix-core", "target","*")
-
-    global hadoop_conf
```
stoty, line 135: see comment for queryserver

bin/queryserver.py
infraio: But the hadoop_conf is only used to start a hbase cluster? For queryserver, it only need to know the hbase cluster, such as a hbase client, which don't need to know the hdfs config? I read queryserver's doc and start queryserver which only have hbase's config. I miss something?

stoty: No, the Hadoop config is also is also used by all Hadoop clients, i.e. anything that uses HDFS, MR, etc.

HBase is designed to work on top of Hadoop, and uses many of its features. It needs the Hadoop config even for its clients.

One example is the proxyuser settings that are defined in the Hadoop config files, and HBase and Phoenix both need to access those.

The Phoenix client also uses HDFS directly (at least for loading UDF jars), so its needs to have access to all HDFS settings as well.

I am sure that there are plenty of other cases where HBase (client) and Phoenix need to have access to the full Hadoop config, the above two are just the ones that have caused problems for me personally.

infraio: Got it. Thanks for explanation.

PR #36, opened 2020-06-04
PHOENIX-5936 sqlAlchemy get_columns KeyError: None

stoty:


stoty:

ColumnsRequest is present in the protobuf generated python code, but is not exposed through the PythonDB connection object.

Looks like PEP-0249 doesn't even define table metadata operations, though we could certainly expose it it via a non-standard method.

PR #37, opened 2020-06-05
PHOENIX-5938 Support impersonation in the python driver

stoty:


python/phoenixdb/phoenixdb/__init__.py
```diff
@@ -101,33 +111,65 @@ def connect(url, max_retries=None, auth=None, authentication=None, avatica_user=
...
         del url_params['truststore']
-        rebuild = True
-
-    if rebuild:
-        url_parsed._replace(query=urlencode(url_params, True))
+        params_changed = True
+
+    if authentication == 'BASIC' or authentication == 'DIGEST':
+        # Handle standard user and password parameters
+        if user is not None and avatica_user is None:
```
joshelser, line 157: As a meta-comment, I don't know why I felt the need to distinguish the 'avatica_user' from the normal 'user'. I think I had some idea where PQS could provide some authz and then the database has its own? I wonder if this is just too complicated.

stoty: That would make sense, as Avatica is a kind of proxy. Its users don't seem to need that though.
We could add similar alias options to the JDBC driver, maybe.

PR #39, opened 2020-06-16
PHOENIX-5901 Fix licensing in preparation for first python release

joshelser:
Use RAT to report on our source code. RAT on the CLI apparently can't
exclude things based on path, so the protobuf-generated code was
manually edited.


python/phoenixdb/NEWS.rst
```diff
@@ -1,3 +1,19 @@
+..
```
stoty, line 1: According to http://www.apache.org/legal/src-headers.html#faq-exceptions README type .rst or .RST files don't need a license header. 
Since these files are meant to be read without preprocessing, the license headers are inconvenient for the reader. (I am not talking about the .rst files in the doc dir, which are meant to be processed)

python/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
stoty, line 1: Do we need license files in this top level directory, when we have the licenses one level down together with the real content ? 

I also think that this directory should only contain the phoenixdb subdirectory, and we should remove the README.md file. It's redundant, and since it's not in the root directory of a git repo, it's not very useful in the context of github either.

Perhaps we could just flatten the directory structure, and instead of /python/phoenixdb/stuff, we could have just python-phoenixdb/stuff ?

python/phoenixdb/NEWS.rst
joshelser: Hrm, I would normally disagree with you, but if I read, alone, "Short informational text files", I can see your interpretation of this. I'll back it out of these three files (NEWS, README, and RELEASING) and add rat check exclusions.

python/LICENSE
joshelser: I put them in the python/ as a guess of what we would include in the src-release. The (strict) requirement is that our source release should have LICENSE and NOTICE files in the root of the source release artifact(s) (i.e. the root level of the zip or tarball).

The guidance here http://www.apache.org/legal/src-headers.html#notice speaks in terms of the source release. I don't remember for sure if that also applies to the "root of the source code repository". Convention is that most projects have an L&N file at the root of the repo.

> Do we need license files in this top level directory, when we have the licenses one level down together with the real content ?

The L&N for the queryserver release will be different than the one we have for python. Specifically, all of the stuff in NOTICE _should not_ appear in the queryserver's release NOTICE. So, it would be good, IMO, whatever we consider the "root" of phoenixdb have the L&N there.

> I also think that this directory should only contain the phoenixdb subdirectory, and we should remove the README.md file. It's redundant, and since it's not in the root directory of a git repo, it's not very useful in the context of github either.

Agree.

> Perhaps we could just flatten the directory structure, and instead of /python/phoenixdb/stuff, we could have just python-phoenixdb/stuff ?

Agree.

joshelser:
> I've just noticed that this doesn't touch the LICENSE and NOTICE files in the python/phoenixdb directory.

I don't see any L&N files in python/phoenixdb. Are you talking about the LICENSE files in the .eggs?

joshelser:
c6da98d and f389a24 should address the comments from your last review, Istvan. Thanks for looking.

stoty:
> > I've just noticed that this doesn't touch the LICENSE and NOTICE files in the python/phoenixdb directory.
> 
> I don't see any L&N files in python/phoenixdb. Are you talking about the LICENSE files in the .eggs?

My mistake, I saw them locally on my system.

There are two issues left:

* We need to update the python script path for the java tests.
(SecureQueryServerPhoenixDBIT SecureQueryServerPhoenixDBIT.java lines 342 and 367, and test_phoenixdb.sh line 63)

* May want to consider removing README.md. Looks like github will show README.rst the same way if it can parse it. (you seemed to agree to do it, but maybe you only meant the directory flattening)  

And a bonus one, for maybe another ticket: Why doesn't mvn install run the ITs ?

joshelser:
> We need to update the python script path for the java tests.
(SecureQueryServerPhoenixDBIT SecureQueryServerPhoenixDBIT.java lines 342 and 367, and test_phoenixdb.sh line 63)

Oh, good catch :)

> May want to consider removing README.md. Looks like github will show README.rst the same way if it can parse it. (you seemed to agree to do it, but maybe you only meant the directory flattening)

Oops, yeah, I meant to remove the markdown readme (rather, consolidate with the rst readme)

stoty:
This will need a rebase because of PHOENIX-5964. 
Would you like me to do the rebase and the agreed remaining changes ?

joshelser:
I had the code changes staged here and (I think) got distracted waiting for a test result and then just never pushed.

If this is holding you up, please feel free to rebase it and go ahead with it tmrw. Otherwise, I'll try to do the rebase myself first thing :)

joshelser:
I'm not seeing the stdout/stderr from the src/it/bin script in the test log. I'm not sure if I'm looking in the correct place, or what the deal is. Do you see that normally, Istvan?

joshelser:
```
[INFO] Running org.apache.phoenix.end2end.SecureQueryServerPhoenixDBIT
[WARNING] Tests run: 3, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 431.013 s - in org.apache.phoenix.end2end.SecureQueryServerPhoenixDBIT
```

Finally! Maybe the `pip install` path fixed it?

stoty:
+1 LGTM

I think that I did see the output of the scripts previously, but honestly I was more concerned with letting the scripts finish, rather than hanging indefinitely with a full stdout/stderr buffer, than with the output :)

Anyway, your solution is much cleaner than relying on the JVM exit cleanups.

I've run built it with both default settings,  and the full python test suite, and everything looks great.

PR #40, opened 2020-06-29
PHOENIX-5964 Rename queryserver subprojects

stoty:


stoty:

PR #42, opened 2020-07-07
PHOENIX-5994 SqlAlchemy schema filtering incorrect semantics

stoty:


python-phoenixdb/phoenixdb/tests/test_sqlalchemy.py
```diff
@@ -52,6 +52,39 @@ def test_textual(self):
...
+                connection.execute('drop table if exists A.ALCHEMY_TEST_A')
+                connection.execute('drop table if exists B.ALCHEMY_TEST_B')
+
+                connection.execute(text('create table ALCHEMY_TEST (ID integer primary key)'))
+                connection.execute(text('create table A.ALCHEMY_TEST_A (ID_A integer primary key)'))
+                connection.execute(text('create table B.ALCHEMY_TEST_B (ID_B integer primary key)'))
+
+                inspector = db.inspect(engine)
+
+                self.assertEqual(inspector.get_schema_names(), [None, 'A', 'B', 'SYSTEM'])
```
joshelser, line 69: Is returning `None` vs `''` (empty string) intentional? i.e. is this convention for sqlalchemy?

stoty: I am not actually sure. The documentation doesn't specify this.

The internal representation in Phoenix is null, which translates to None.
When specifying the schema filter, the default schema is also represented in sqlAlchemy by None.

Based on this, returning None here seems to make sense. 

I'll try to figure out what i.e. the mysql driver returns here, and adjust if it returns ''

joshelser: > The internal representation in Phoenix is null, which translates to None.

So, Phoenix is really weird when it comes to "null" and "empty string", in that they are functionally equivalent. Not sure if that's a surprise to you.

That all said, `None` is fine. Mostly curious :)

stoty: In most DBs the default scheme seems to have textual representation, so I couldn't find anything definitive by a cursory search. 

At least we also return None from **sqlalchemy.engine.reflection.Inspector.default_schema_name** , so we are internally consistent.

I'll try to check how Hue handles this semantics tomorrow. (Should have done that before opening the PR)

I've run into the null = '' weirdness before, but it seems to work in direction only. (i.e. `column = ''` won't match on empty strings / null , only `is null` will)

stoty:
I've extensively rewritten the MetaData handling (and some other components)
Please check the new code as well.

This version (hopefully) consistently uses the empty string to indicate the default schema, catalog, etc.

python-phoenixdb/phoenixdb/connection.py
```diff
@@ -135,50 +158,51 @@ def set_session(self, autocommit=None, readonly=None):
...
+        self._avatica_props = self._client.connection_sync_dict(self._id, {'transactionIsolation': bool(value)})
+
+    def meta(self):
+        """Creates a new meta.
+
+        :returns:
+            A :class:`~phoenixdb.meta` object.
+        """
+        if self._closed:
+            raise ProgrammingError('the connection is already closed')
```
joshelser, line 203: nit: make it a complete sentence, `The .. already closed.`

python-phoenixdb/phoenixdb/sqlalchemy_phoenix.py
```diff
@@ -126,76 +126,65 @@ def create_connect_args(self, url):
...
+        pk_columns = [col['COLUMN_NAME'] for col in columns if col['KEY_SEQ'] > 0]
+        return {'constrained_columns': pk_columns}
+
+    def get_indexes(self, conn, table_name, schema=None, **kw):
+        '''This information does not seem to be exposed via Avatica
+        TODO: Implement by directly querying SYSTEM tables ? '''
         return []
 
     def get_foreign_keys(self, conn, table_name, schema=None, **kw):
+        '''Foreign keys are a foreign concept to Phoenix'''
```
joshelser, line 176: Should we throw some kind of NotImplemented Exception?

python-phoenixdb/phoenixdb/avatica/client.py
```diff
@@ -301,18 +313,26 @@ def connection_sync(self, connection_id, connProps=None):
...
         request.connection_id = connection_id
-        request.conn_props.auto_commit = connProps.get('autoCommit', False)
         request.conn_props.has_auto_commit = True
-        request.conn_props.read_only = connProps.get('readOnly', False)
         request.conn_props.has_read_only = True
-        request.conn_props.transaction_isolation = connProps.get('transactionIsolation', 0)
-        request.conn_props.catalog = connProps.get('catalog', '')
-        request.conn_props.schema = connProps.get('schema', '')
+        if 'autoCommit' in connProps:
+            request.conn_props.auto_commit = connProps.pop('autoCommit')
```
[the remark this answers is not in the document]
stoty: That's a good idea, the caller may want to reuse a connProps instance.

python-phoenixdb/phoenixdb/connection.py
stoty: Sure. I'll search and replace it everywhere.

python-phoenixdb/phoenixdb/sqlalchemy_phoenix.py
stoty: That doesn't work. 

SqlAlchemy assumes that if we implement _get_pk_constraint()_ we also implement  _get_foreign_keys()_, and fails to initialize its Table object without it.

joshelser: ah, that's fun! :P

joshelser:
`0bafa1b` looks great. Ship it!

PR #43, opened 2020-07-10
PHOENIX-5999 Hack together a better implementation for executemany us…

joshelser:
…ing ExecuteBatch

Need to get tests updated too and make sure this is adapted to sqlalchemy bindings as well, but sharing now.

### DIFF
--- a/python-phoenixdb/phoenixdb/avatica/client.py
+++ b/python-phoenixdb/phoenixdb/avatica/client.py
@@ -480,6 +480,39 @@ class AvaticaClient(object):
         response.ParseFromString(response_data)
         return response.results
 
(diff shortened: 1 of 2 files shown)


stoty:
Looks good. (but I agree that needs a test)

Actually, I don't think we need (or indeed can) do anything in the SqlAlchemy driver for this, this seems to be handled in SqlAlchemy core.

stoty:
We already have executemany() tests from dbapi20.py, so that angle is covered as well.

The only changes that I've made is updating the exception message (and all remaining exception messages) to be full sentences, as you suggested in a previous review.

joshelser:
Thanks Istvan!

PR #45, opened 2020-07-23
PHOENIX-6006 Bump queryserver version to 6.0

stoty:


joshelser:
My original idea in making these be 1.0.0 was that it clearly separates it from the Phoenix versioning. I don't think we want to relate phoenix release versions with phoenix queryserver release versions.

I have two concerns with using 6.0.0. One is that we'll have confusion when Phoenix inevitably wants to also use 6.0.0. The other is that a version which is "close" to the core Phoenix codebase might imply that this has tie-ins with Phoenix (e.g. "I have to use a verison of phoenix-queryserver which matches my phoenix version").

My hope was that we could just always have the "latest" phoenix-queryserver release which people can pick up, and overlay the appropriate phoenix-client jar onto it.

stoty:
These are valid concerns @joshelser . 

However, we already have these artifacts with 4.x releases on maven central, and I am quite sure that most ppl would download 5.0 instead of, say 1.1.3 when both are available.
https://mvnrepository.com/artifact/org.apache.phoenix/phoenix-queryserver-client

if we stick to 1.0, I think that we should make sure that the maven **artifactId**s do not match the corresponding old 4.x and 5.0 artifacts, otherwise those old artifacts will always appear newer than the new ones, and users will keep running into trouble. 

We may have to change/undo some of the recent queryserver artifact renaming, and I think that we'll have the same problem with connectors as well.

joshelser:
> if we stick to 1.0, I think that we should make sure that the maven artifactIds do not match the corresponding old 4.x and 5.0 artifacts, otherwise those old artifacts will always appear newer than the new ones, and users will keep running into trouble.

Oh, really good point. I hadn't thought about that.

Let's just stick with 6.0.0 then. We can update website documentation so that people know how we're managing these versions (and the normal phoenix versions as well as phoenix-connectors :P)

PR #46, opened 2020-07-29
PHOENIX-5829 Make it possible to build/test queryserver against all s…

stoty:
…upported versions

also fix some regressions in python scripts and assembly


phoenix-queryserver-it/pom.xml
```diff
@@ -213,5 +175,92 @@
...
+  <profiles>
+    <!-- Phoenix 4.16.0-SNAPSHOT doesn't need special handling -->
+    <profile>
+      <id>phoenix-5.1.0-SNAPSHOT</id>
+      <activation>
+        <property>
+          <name>phoenix.version</name>
+          <value>5.1.0-SNAPSHOT</value>
+        </property>
+      </activation>
```
joshelser, line 194: Is there a reason to only have the activation for 5.1.0-SNAPSHOT?

README.md
```diff
@@ -44,14 +44,12 @@ $ mvn package
...
-and specify properties to indicate a specific Phoenix version.
+To build a release of PQS which packages a specific version of Phoenix, specify the `package-phoenix-client` system property
+and specify the `phoenix.version` system property to indicate a specific Phoenix version.
 
-By default, PQS will package the same version of Phoenix used for build/test. This version is controlled by the system
-property `phoenix.version` system property. Depending on the version of Phoenix, you may also be required to
-use the `phoenix.hbase.classifier` system property to identify the correct version of Phoenix built against
-the version of HBase of your choosing.
+PQS will package the same version of Phoenix used for build/test. This version is controlled by the
+ `phoenix.version` system property.
```
joshelser, line 51: If we don't have activation for all profiles off of phoenix.version values, we should call out the need to explicitly provide the profile.

stoty: It's a good idea, that way we document how to built with future/custom versions.

phoenix-queryserver-it/pom.xml
stoty: 5.0.0 includes PQS, and 5.1.0 is not yet released.
I was planning to add 5.1.0 after it is released.
However, I am not against adding either or both to the activation criteria now.

joshelser: It just seemed strange to have activation for one of them but not the others. I think we should just be consistent in one direction :)

stoty:
I've removed the automatic profile activation, and added the information to the readme instead, as you suggested.

PR #47, opened 2020-08-04
PHOENIX-6060 Create release script for python-phoenixdb ASF releases

stoty:


python-phoenixdb/dev-support/make_rc.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "Script that assembles all you need to make an RC."
+echo "It generates source tar in release directory"
+echo "Presumes that you can sign a release as described at https://www.apache.org/dev/release-signing.html"
+echo "Starting...";sleep 2s
+
+# Set directory variables
+DIR_ROOT="$(cd $(dirname $0);pwd)/.."
+echo $DIR_ROOT
+cd $DIR_ROOT
+VERSION=$(cat setup.py | grep '^version = ".*"$' | grep -o '".*"' |sed 's/"//g')
```
joshelser, line 32: nit just `grep '^version = ".*"$' setup.py`

python-phoenixdb/dev-support/make_rc.sh
```diff
@@ -0,0 +1,91 @@
...
+DIR_REL_BASE=$DIR_ROOT/release
+echo $DIR_REL_BASE
+DIR_REL_ROOT=$DIR_REL_BASE/python-phoenixdb-$VERSION
+echo $DIR_REL_ROOT
+REL_SRC=python-phoenixdb-$VERSION-src
+echo $REL_SRC
+DIR_REL_SRC_TAR_PATH=$DIR_REL_ROOT/src
+echo $DIR_REL_SRC_TAR_PATH
+
+git clean -fx .
```
joshelser, line 44: Warn people that this will overwrite changes in the entire repository first?

python-phoenixdb/dev-support/make_rc.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "Presumes that you can sign a release as described at https://www.apache.org/dev/release-signing.html"
+echo "Starting...";sleep 2s
+
+# Set directory variables
+DIR_ROOT="$(cd $(dirname $0);pwd)/.."
+echo $DIR_ROOT
+cd $DIR_ROOT
+VERSION=$(cat setup.py | grep '^version = ".*"$' | grep -o '".*"' |sed 's/"//g')
+
+echo $VERSION
```
joshelser, line 34: Some labels to accompany these variables (this `echo` and the following 4) would be very appreciated.

python-phoenixdb/dev-support/make_rc.sh
```diff
@@ -0,0 +1,91 @@
...
+DIR_REL_SRC_TAR_PATH=$DIR_REL_ROOT/src
+echo $DIR_REL_SRC_TAR_PATH
+
+git clean -fx .
+
+# Generate src tar
+ln -s . $REL_SRC; tar cvzf $REL_SRC.tar.gz --exclude="$REL_SRC/$REL_SRC" $REL_SRC/*; rm $REL_SRC;
+
+# Generate directory structure
+mkdir $DIR_REL_BASE;
```
joshelser, line 50: What if the directory already exists? Do you want to remove it, or is it safe to keep around?

python-phoenixdb/dev-support/make_rc.sh
stoty: Done.

python-phoenixdb/dev-support/make_rc.sh
stoty: Good idea. Done.

python-phoenixdb/dev-support/make_rc.sh
stoty: Those are I just debug echos that I forgot to remove.
I don't think they are needed in production (phoenix core doesn't print them either)

python-phoenixdb/dev-support/make_rc.sh
stoty: git clean -fx will nuke it, so it's not an issue.

PR #49, opened 2020-09-15
PHOENIX-5869 Use symlinks to reduce size of phoenix queryserver assembly

stoty:
also fix top directory name in assmebly


richardantal:
+1
This change looks good to me, I have also tried it and it replaces the phoenix-queryserver-client .jar as we would like to.

PR #53, opened 2021-01-04
PHOENIX-6294 javax.servlet relocation added by PHOENIX-6151 breaks PQS

stoty:


stoty:
+1 LGTM
I have tested the patch with 4.15 and 5.1.0-SNAPSHOT, and it fixes the issues without breaking 4.15

joshelser (approved):
Thanks! +1 -- this is what I did locally and it did the trick.

PR #55, opened 2021-02-03
PHOENIX-6177 Queryserver Avatica and Jetty versions are incompatible

stoty:
sync jetty version to avatica 1.17.0


joshelser (approved):
Good catch. Let's drive an Avatica release so we can undo this sooner than later :)

I'll see if I can work this in after I get the spnego improvements done.

stoty:
Thanks Josh.
IIRC this one is already fixed in Avatica, it just hasn't been released yet.

PR #59, opened 2021-03-26
PHOENIX-6414 Explicit creation of auth mech for SPNEGO

TrilceAC:
This PR applies the patch proposed at [PHOENIX-6414](https://issues.apache.org/jira/browse/PHOENIX-6414) with some minor corrections. It is intended to make Phoenix 5.0.0 (PQS) accessible from Python when SPNEGO is used as the authentication mechanism.


stoty (commented):
+1 LGTM

PR #60, opened 2021-04-30
PHOENIX-6459 Try to use a Session to pass along any cookies the serve…

joshelser:
…r sends

The Requests documentation also alludes to performance benefit due to
leaving the existing connection open (rather than re-connecting each time)

  https://docs.python-requests.org/en/master/user/advanced/#session-objects

### DIFF
--- a/python-phoenixdb/phoenixdb/avatica/client.py
+++ b/python-phoenixdb/phoenixdb/avatica/client.py
@@ -149,7 +149,7 @@ class AvaticaClient(object):
         self.auth = auth
         self.verify = verify
         self.connection = None
-        self.session = requests.Session()
+        self.session = None
 
     def connect(self):
         """This method used to open a persistent TCP connection
@@ -161,15 +161,23 @@ class AvaticaClient(object):
         pass
 
     def _post_request(self, body, headers):
+        # Create the session if we haven't before
+        if not self.session:
+            logger.debug("Creating a new Session")
+            self.session = requests.Session()
+            self.session.headers.update(headers)
+            self.session.stream = True
+            if self.auth is not None:
+                self.session.auth = self.auth
+
         retry_count = self.max_retries
         while True:
-            logger.debug("POST %s %r %r", self.url.geturl(), body, headers)
+            logger.debug("POST %s %r %r", self.url.geturl(), body, self.session.headers)
 
-            requestArgs = {'data': body, 'stream': True, 'headers': headers}
+            requestArgs = {'data': body}
 
-            if self.auth is not None:
-                requestArgs.update(auth=self.auth)
-
+            # Setting verify on the Session is not the same as setting it
+            # as a request arg
             if self.verify is not None:
                 requestArgs.update(verify=self.verify)
 
--- a/python-phoenixdb/phoenixdb/avatica/client.py
+++ b/python-phoenixdb/phoenixdb/avatica/client.py
@@ -149,6 +149,7 @@ class AvaticaClient(object):
         self.auth = auth
         self.verify = verify
         self.connection = None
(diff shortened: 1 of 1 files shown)


python-phoenixdb/phoenixdb/avatica/client.py
```diff
@@ -149,6 +149,7 @@ def __init__(self, url, max_retries=None, auth=None, verify=None):
         self.auth = auth
         self.verify = verify
         self.connection = None
+        self.session = requests.Session()
```
stoty, line 152: IMO this would better in open_connection()

python-phoenixdb/phoenixdb/avatica/client.py
```diff
@@ -173,7 +174,7 @@ def _post_request(self, body, headers):
                 requestArgs.update(verify=self.verify)
```
stoty, line 174: Wouldn't performing the authentication setup  on the Session object in open_connection() once per sessions be better?

We could also pre-set the content-type on the Session object, and wouldn't have to pass the headers parameter around.

The same could be done for the stream flag.

https://docs.python-requests.org/en/master/user/advanced/#session-objects
https://2.python-requests.org/en/master/api/#request-sessions

stoty:
This may also help performance by pooling the underlying HTTP connection ?

python-phoenixdb/phoenixdb/avatica/client.py
joshelser: I would guess that we wouldn't see a big improvement while we are suffering from CALCITE-4152 (spnego re-negotiation), however I agree with you.

I'll try to implement your suggestions. Thanks!

python-phoenixdb/phoenixdb/avatica/client.py
joshelser: actually, I've pushed this down into `_post_request` to avoid having to create a bunch of "if not self.session` checks.

joshelser:
```
global
  maxconn 256
defaults
  mode http
  timeout connect 5000ms
  timeout client 60000ms
  timeout server 60000ms
frontend http-in
  bind  *:8080
  default_backend servers
backend servers
  balance roundrobin
  cookie PQS-SESSION insert indirect nocache
  server queryserver1 localhost:8765 check cookie queryserver1
  server queryserver2 localhost:8766 check cookie queryserver2
```

Used the above to validate that this actually works against my local machine with a basic `select * from TABLE` in Python

stoty (approved):
+1 LGTM

joshelser:
> This may also help performance by pooling the underlying HTTP connection ?

Yeah, it might. I haven't done a test to verify that yet. I should be able to pretty easy. Let me see :)

joshelser:
```
start = time.perf_counter_ns()
iterations = 500
# each connection results in a new Session
with db.connect() as conn:
    for x in range(1, iterations):
        with conn.cursor() as cursor:
            cursor.execute("SELECT count(1) FROM employees limit 10")
            row=cursor.fetchone()
        if x % 100 == 0:
            print("Iteration %d done" % (x))

end = time.perf_counter_ns()
print("Read {} rows, Total execution time {} milliseconds".format(iterations, (end - start) / 1000 / 1000))
```

Using this and going directly against PQS without and with this change does show some interesting results!

The query here is running `SELECT count(1) from EMPLOYEES` which has 50k rows in it. Small data size, fully cached in HBase, trying to eliminate any latency from reading the data itself.

After warming up the JVM, I ran 3 executions with and without this change and averaged the results. Without this change, average execution time was 36673ms. With this change, the execution time was 31799ms. That's about a 15% performance improvement!

PR #64, opened 2021-05-17
PHOENIX-6410 Expose primary keys metadata via SqlAlchemy

stoty:


romainr:
I can see for pip installing your branch and testing within Hue

romainr:
Nice!

image

python-phoenixdb/phoenixdb/tests/test_db.py
```diff
@@ -172,6 +172,66 @@ def test_meta(self):
...
+                    self.assertTrue(False)
+                except ProgrammingError:
+                    pass
+
+                self.maxDiff = None
+
+                self.assertEqual(meta.get_index_info(table='NON_EXISTENT'), [])
+
+                self.assertTrue(len(meta.get_index_info(table='DEFAULT_TABLE')) > 1)
+                print(meta.get_index_info(table='DEFAULT_TABLE'))
```
romainr, line 233: nit: print

romainr (approved):
Also confirming we can get the indexes

`[{'name': 'GLOBAL_IDX', 'unique': True, 'column_names': ['0:MESSAGE', ':ID']}]`

python-phoenixdb/phoenixdb/tests/test_db.py
stoty: THX, done.

stoty:
I have also added some post-processing in the SQLAlchemy index column list to get the expected name when there is no column family trickery.

joshelser (commented):
Belated +1 from me :)

PR #65, opened 2021-05-19
PHOENIX-6473 Add Hadoop JMXServlet as /jmx endpoint

anmolnar:
Added new Avatica server customizer to create `/jmx` endpoint powered by JMXServlet. Also had to add jetty-servlet dependency and remove shading of jetty, otherwise the Servlet class cannot be instantiated.

### DIFF
--- a/phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/JMXJsonEndpointServerCustomizer.java
+++ b/phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/JMXJsonEndpointServerCustomizer.java
@@ -1,0 +1,41 @@
+package org.apache.phoenix.queryserver.server.customizers;
+
+import org.apache.calcite.avatica.server.ServerCustomizer;
+import org.apache.hadoop.jmx.JMXJsonServlet;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.HandlerList;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.webapp.WebAppContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+public class JMXJsonEndpointServerCustomizer implements ServerCustomizer<Server> {
+  private static final Logger LOG = LoggerFactory.getLogger(JMXJsonEndpointServerCustomizer.class);
+
+  @Override
+  public void customize(Server server) {
+    Handler[] handlers = server.getHandlers();
+    if (handlers.length != 1) {
+      LOG.warn("Observed handlers on server {}", Arrays.toString(handlers));
+      throw new IllegalStateException("Expected to find one handler");
+    }
+    HandlerList list = (HandlerList) handlers[0];
+
+    WebAppContext ctx = new WebAppContext();
+    ctx.setContextPath("/");
+    ServletHolder holder = new ServletHolder(JMXJsonServlet.class);
+    ctx.addServlet(holder, "/jmx");
+
+    ctx.setServer(server);
+
+    Handler[] realHandlers = list.getChildHandlers();
+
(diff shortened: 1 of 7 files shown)


anmolnar:
@stoty @richardantal PTAL.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/JMXJsonEndpointServerCustomizer.java
```diff
@@ -0,0 +1,46 @@
+package org.apache.phoenix.queryserver.server.customizers;
```
joshelser, line 1: Needs a license header

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/JMXJsonEndpointServerCustomizer.java
```diff
@@ -0,0 +1,46 @@
...
+
+    Servlet servlet = new JMXJsonServlet();
+    ServletHolder holder = new ServletHolder(servlet);
+    ctx.addServlet(holder, "/");
+
+    Handler[] realHandlers = list.getChildHandlers();
+    Handler[] newHandlers = new Handler[realHandlers.length + 1];
+    newHandlers[0] = ctx;
+    System.arraycopy(realHandlers, 0, newHandlers, 1, realHandlers.length);
+    server.setHandler(new HandlerList(newHandlers));
```
joshelser, line 44: Does this end up requiring authentication to access it? I don't remember if we do authentication at the Server level or in the PQS Handler itself.

joshelser (changes requested):
> Also had to add jetty-servlet dependency and remove shading of jetty, otherwise the Servlet class cannot be instantiated.

What kind of error did you see when jetty-servlet is shaded? More curious than concerned (given PQS is standalone).

Need to fix the license headers, otherwise, this looks like a great addition!

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/JMXJsonEndpointServerCustomizer.java
anmolnar: No, it doesn't need authentication (currently).

anmolnar:
@joshelser This is the error that I see if I don't touch relocations:
```
21/05/20 17:31:58 ERROR server.QueryServer: Unrecoverable service error. Shutting down.
java.lang.NoClassDefFoundError: javax/servlet/Servlet
        at org.apache.phoenix.queryserver.server.ServerCustomizersFactory$ServerCustomizersFactoryImpl.createServerCustomizers(ServerCustomizersFactory.java:75)
```

anmolnar:
@stoty suggested to remove the entire relocations part which caused another runtime problem:
```
javax.servlet.UnavailableException: Servlet class org.apache.hadoop.jmx.JMXJsonServlet is not a javax.servlet.Servlet
	at org.eclipse.jetty.servlet.ServletHolder.checkServletType(ServletHolder.java:555)
	at org.eclipse.jetty.servlet.ServletHolder.doStart(ServletHolder.java:385)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at org.eclipse.jetty.servlet.ServletHandler.lambda$initialize$0(ServletHandler.java:749)
```

I also wanted to add an integration test to validate the new endpoint, but I see the above error no matter how I tweak relocation. To be honest shading is completely jibbrish to me, so I've given up adding a test.

stoty:
I tried to figure out if the Jmx servlet is authenticatd in Hadoop, but I got lost in the code.
Did you check if you can access Jmx without authentication in Hadoop ?
If yes, then it's fine, if not then, then we should also require authentication here.

Also, I think that a property to optionally disable the JMX server would be useful.

joshelser:
I forgot to reply last week, but I think this is fine to leave it unauthenticated. I can't think of anything harmful exposed via the servlet. Thanks for adding the option to disable.

joshelser (approved):

PR #70, opened 2021-06-16
PHOENIX-6003 Metadata operations via Avatica turn empty string args t…

stoty:
…o null

Rebuild phoenixdb with avatica 1.18.0 protobuf, and use new metadata
call fields


stoty:
Tests fail because of PHOENIX-6493
Will not commit until after we've fixed PHOENIX-6493, and updated the default phoenix release in PQS to one with the fix.

PR #71, opened 2021-07-07
PHOENIX-6407 phoenixdb for Python silently ignores placeholders < pla…

richardantal:
…ceholder arguments

Change-Id: I1715aca93420686585d5c6d4db7994c74bd39cba

### DIFF
--- a/python-phoenixdb/phoenixdb/cursor.py
+++ b/python-phoenixdb/phoenixdb/cursor.py
@@ -174,6 +174,8 @@ class Cursor(object):
             return self._process_result(results[0])
 
     def _transform_parameters(self, parameters):
+        if len(parameters) != len(self._parameter_data_types):
+            raise ProgrammingError('Number of placeholders (?) must match number of parameters. Number of placeholders: {0}. Number of parameters: {1}'.format(len(self._parameter_data_types), len(parameters)))
         typed_parameters = []
         for value, data_type in zip(parameters, self._parameter_data_types):
             field_name, rep, mutate_to, cast_from, is_array = data_type
--- a/python-phoenixdb/phoenixdb/tests/test_db.py
+++ b/python-phoenixdb/phoenixdb/tests/test_db.py
@@ -285,3 +285,21 @@ class PhoenixDatabaseTest(DatabaseTestCase):
                 cursor.execute('drop table if exists AAA')
                 cursor.execute('drop table if exists "aaa"')
                 cursor.execute('drop table if exists "Aaa"')
+
+    def test_param_number_mismatch(self):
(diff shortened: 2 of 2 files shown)


stoty:
I think that this a problem with either Phoenix or perhaps Avatica.
I don't think that this should be fixed in phoenixdb.

stoty:
I'd expect this to be reproducibale even with the Java thick client.

richardantal:
I think this ticket is about calling the cursor.execute with mismatching number of placeholders in phonixdb.

I created a table 
`cursor.execute("CREATE TABLE users (id INTEGER PRIMARY KEY, username VARCHAR, name VARCHAR)")`
and ran these upserts
```
cursor.execute("UPSERT INTO users VALUES (?, ?, ?)", (1, '1', '1'))
cursor.execute("UPSERT INTO users VALUES (?, ?, ?)", (2, '2', '2', '2'))
cursor.execute("UPSERT INTO users VALUES (?, ?, ?)", (3, '3'))
cursor.execute("UPSERT INTO users VALUES (?, ?, ?, ?)", (4, '4', '4'))
cursor.execute("UPSERT INTO users VALUES (?, ?)", (5, '5', '5'))

cursor.execute("UPSERT INTO users VALUES (?, ?)", (6, '6'))
```

Without this change:
1, 2, 5 and 6 was successful.
the result of the select:
```
`[[1, '1', '1'], [2, '2', '2'], [5, '5', None], [6, '6', None]]`
3 -> InternalError: ('Parameter value unbound. Parameter 3 is unbound', 2004, 'INT05', None)
4 -> ProgrammingError: ('Number of columns upserting must match number of values. Numbers of columns: 3. Number of values: 4 tableName=USERS', 1020, '42Y60', None)
```

With my change
only 1 and 6 is successful
```
2 -> ProgrammingError: ('Number of placeholders (?) must match number of parameters. Number of placeholders: 3. Number of parameters: 4', None, None, None)
3 -> ProgrammingError: ('Number of placeholders (?) must match number of parameters. Number of placeholders: 3. Number of parameters: 2', None, None, None)
4 -> ProgrammingError: ('Number of columns upserting must match number of values. Numbers of columns: 3. Number of values: 4 tableName=USERS', 1020, '42Y60', None)
5 -> ProgrammingError: ('Number of placeholders (?) must match number of parameters. Number of placeholders: 2. Number of parameters: 3', None, None, None)
```

richardantal:
If we have more parameter than placeholders the zip will eliminate the extra ones here:
https://github.com/apache/phoenix-queryserver/blob/master/python-phoenixdb/phoenixdb/cursor.py#L178
Thats why it is silently ignored.

joshelser (approved):
Looks great to me. I could see an argument for requiring at least as many values as you have placeholders, but I think requiring them to be equal in length is great.

PR #80, opened 2021-12-16
PHOENIX-6613 QueryServerBasicsIT usage by external tools

vdiravka:
QueryServerBasicsIT is a public class which can allow testing PQS in external tools.
It is possible to setup PQS, but there is no possibility to connect to it withDriverManager.getConnection(CONN_STRING), because there is no getter for CONN_STRING.


stoty (changes requested):
Generally OK, but please revert the too long lines.

stoty:
I've committed the change without the changes that result in too long lines.

PR #83, opened 2022-03-02
PHOENIX-6654 queryserver.py sets the umask to 0000 when starting PQS …

stoty:
…in daemon mode


chrajeshbabu (approved):

joshelser (approved):
Fun API from python (no getter for the current umask).

LGTM

PR #100, opened 2022-07-13
PHOENIX-6745 Fix run-source-ratcheck.sh script

richardantal:
Change-Id: I05ea9eaeb53510bbd276f7b3925abd6b329af3ee


richardantal:
We could release version 1.2.0 phoenixDB without this change this is not a blocker.
It is totally fine to merge after the release.

stoty:
See my comment on the ticket.

PR #101, opened 2022-08-08
PHOENIX-6762 Phoenix QueryServer cannot run correctly with python 3.8+

stoty:
backport https://pagure.io/python-daemon/c/b708912

### DIFF
--- a/bin/daemon.py
+++ b/bin/daemon.py
@@ -38,6 +38,7 @@
 #
 # Apache Phoenix note: this file is `daemon.py` from the package
 # `python-daemon 2.0.5`, https://pypi.python.org/pypi/python-daemon/
+# It has been modified in PHOENIX-6762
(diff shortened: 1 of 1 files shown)


bin/daemon.py
```diff
@@ -727,9 +728,9 @@ def is_socket(fd):
         """
     result = False
 
-    file_socket = socket.fromfd(fd, socket.AF_INET, socket.SOCK_RAW)
```
richardantal, line 731: Nit: if you could delete one of the 2 empty lines it would be a bit nicer

richardantal (approved):
LGTM

bin/daemon.py
stoty: Thanks, fixed.
Also reworeded the comment at the top.

PR #107, opened 2022-10-03
PHOENIX-6801 Fix failing docker build because of missing files

richardantal:


stoty (approved):
+1 LGTM

However, AFAICT the build is not failing because of missing repos, but because of missing files from the repos. (Which is standard Ubuntu procedure)

I suggest that you fix the ticket/commit/NEWS.rst text to reflect this before commiting.

PR #108, opened 2022-10-04
PHOENIX-6803 make_rc.sh does not work on mac

richardantal:


richardantal:
Thank you @stoty for the review!

PR #112, opened 2022-11-24
PHOENIX-6836: Enable code coverage reporting to SonarQube in Phoenix-Queryserver

horvathdora:


pom.xml
```diff
@@ -701,6 +702,13 @@
             <name>!skip.code-coverage</name>
         </property>
       </activation>
+      <properties>
+          <sonar.dynamicAnalysis>reuseReports</sonar.dynamicAnalysis>
+          <sonar.surefire.reportsPath>${project.build.directory}/surefire-reports</sonar.surefire.reportsPath>
+          <sonar.coverage.jacoco.xmlReportPaths>
+              ${project.build.directory}/site/jacoco/jacoco.xml
```
richardantal, line 709: Why don't we have `phoenix-queryserver-assembly` here like for the phoenix main repo?

dev/code-coverage/run-coverage.sh
```diff
@@ -0,0 +1,70 @@
...
+  SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+  MAIN_POM="${SCRIPT_DIR}/../../pom.xml"
+  # Check the syntax for the THREAD_COUNT variable
+  if [[ "$THREAD_COUNT" =~ ^[0-9]+([.][0-9]+)?$ ]] || [[ "$THREAD_COUNT" =~ ^[0-9]+([.][0-9]+)+[C]?$ ]]; then
+    THREADS="${THREAD_COUNT}"
+  else
+    THREADS=1
+  fi
+
+  mvn -B -e -f "$MAIN_POM" clean verify -Pcodecoverage -fn -Dmaven.javadoc.skip=true -DskipShade -T "$THREADS"
```
richardantal, line 42: Shouldn't we have `coverage` profile here?

richardantal (commented):
I am not sure about the following parts, (it doesn't mean they are wrong).
If you think it is correct please provide some details for me.

dev/code-coverage/run-coverage.sh
horvathdora: Good point regarding the new profile I have added but never used :)
I added the sonar analysis to the original codecoverage profile so my coverage profile is not necessary, looks like I forgot to remove it.

pom.xml
horvathdora: If I do it like in the phoenix component and aggregate the coverage in the assembly module then the phoenix-queryserver-orchestrator module is not included in the coverage however I tried to modified the configuration of the jacoco and sonar profiles.

PR #115, opened 2023-01-26
PHOENIX-6861 PQS fails to start because of relocated javax.servlet wi…

stoty:
…th Phoenix 5.2

- make javax.servlet relocation a build time option controlled by a profile
- update default phoenix version to 5.1.3 and HBase to 2.4.12
- add log4j2 logging test dependencies so that tests can be run with HBase 2.5


joshelser (approved):
Yuck! What a mess.

PR #119, opened 2023-02-14
PHOENIX-6879 Prometheus endpoint implementation

lucakovacs:
In this change: 

- I added a new **servlet (/prometheus)** for exposing Prometheus format metrics
- The metrics' **description** can be enabled by `"/prometheus?description=true"` query string
- The metrics can be **filtered** by `"/prometheus?qry=xyz"` query string

### DIFF
--- a/phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/prometheus/PrometheusEndpointServerCustomizer.java
+++ b/phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/prometheus/PrometheusEndpointServerCustomizer.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.queryserver.server.customizers.prometheus;
+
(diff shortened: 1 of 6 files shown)


phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/prometheus/PrometheusEndpointServerCustomizer.java
```diff
@@ -0,0 +1,57 @@
...
+    public void customize(Server server) {
+        Handler[] handlers = server.getHandlers();
+        if (handlers.length != 1) {
+            LOG.warn("Observed handlers on server {}", Arrays.toString(handlers));
+            throw new IllegalStateException("Expected to find one handler");
+        }
+        HandlerList list = (HandlerList) handlers[0];
+
+        ServletContextHandler ctx = new ServletContextHandler();
+        ctx.setContextPath("/prometheus");
```
MitchellJThomas, line 45: Prometheus typically expects and endpoint path to be "/metrics". 

See https://prometheus.io/docs/introduction/first_steps/#:~:text=Prometheus%20expects%20metrics%20to%20be,%3A%2F%2Flocalhost%3A9090%2Fmetrics.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/customizers/prometheus/PrometheusHadoopServlet.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    void writeMetrics(Writer writer, boolean descriptionEnabled, String queryParam)
+            throws IOException {
+        Collection<MetricsRecord> metricRecords = MetricsExportHelper.export();
+        for (MetricsRecord metricsRecord : metricRecords) {
+            for (AbstractMetric metrics : metricsRecord.metrics()) {
+                if (metrics.type() == MetricType.COUNTER || metrics.type() == MetricType.GAUGE) {
+
+                    String key = toPrometheusName(metricsRecord.name(), metrics.name());
```
MitchellJThomas, line 56: Note that this method and its regexp could be called quite often: per metricRecrod, per metric, per metric scrape interval ([default scrape interval is 1 minute](https://github.com/prometheus/prometheus/blob/6fc5305ce904a75f56ca762281c7a1b052f19092/config/config.go#L142)) could add to a bit of CPU time... if I am right, I suggest [memoizing this function](https://dzone.com/articles/java-8-automatic-memoization)

MitchellJThomas (commented):
Some suggestions.  You might also want to add a unit test or two.

PR #124, opened 2023-03-29
PHOENIX-6917: Added support for getting CoulmnLabel rather than the CoulmnName

satyakommula96:
* Added support  for getting the columnLabel (the “as name”) rather than the columnName with a cursor.

Below is the example

```
calcite 💎:sql> select c1 as hello, c2 as world from int_tbl;
+-------------+-------------+
| c1          | c2          |
|-------------+-------------|
| 5           | 0           |
| -123        | 123         |
| 1           | 123456      |
| -123456     | -123456     |
| 10          | 50          |
| -2147483648 | 1           |
| 2147483647  | -2147483648 |
| <null>      | 10          |
| <null>      | <null>      |
| 1           | 1           |
+-------------+-------------+

```
with this patch i'm able to get column_name as column label

```
calcite 💎:sql> select c1 as hello, c2 as world from int_tbl;

+-------------+-------------+
|       HELLO |       WORLD |
|-------------+-------------|
|           5 |           0 |
|        -123 |         123 |
|           1 |      123456 |
|     -123456 |     -123456 |
|          10 |          50 |
| -2147483648 |           1 |
|  2147483647 | -2147483648 |
|             |          10 |
|             |             |
|           1 |           1 |
+-------------+-------------+
```

### DIFF
--- a/python-phoenixdb/phoenixdb/cursor.py
+++ b/python-phoenixdb/phoenixdb/cursor.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 # TODO see note in Cursor.rowcount()
 MAX_INT = 2 ** 64 - 1
 
-ColumnDescription = collections.namedtuple('ColumnDescription', 'name type_code display_size internal_size precision scale null_ok')
+ColumnDescription = collections.namedtuple('ColumnDescription', 'label type_code display_size internal_size precision scale null_ok')
 """Named tuple for representing results from :attr:`Cursor.description`."""
 
 
@@ -115,7 +115,7 @@ class Cursor(object):
         description = []
         for column in self._signature.columns:
             description.append(ColumnDescription(
-                column.column_name,
+                column.label,
                 column.type.name,
                 column.display_size,
                 None,
--- a/python-phoenixdb/phoenixdb/cursor.py
+++ b/python-phoenixdb/phoenixdb/cursor.py
@@ -386,5 +386,5 @@ class DictCursor(Cursor):
         row = super(DictCursor, self)._transform_row(row)
         d = {}
         for ind, val in enumerate(row):
-            d[self._signature.columns[ind].column_name] = val
+            d[self._signature.columns[ind].label] = val
         return d
--- a/python-phoenixdb/phoenixdb/cursor.py
+++ b/python-phoenixdb/phoenixdb/cursor.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 # TODO see note in Cursor.rowcount()
 MAX_INT = 2 ** 64 - 1
 
-ColumnDescription = collections.namedtuple('ColumnDescription', 'label type_code display_size internal_size precision scale null_ok')
+ColumnDescription = collections.namedtuple('ColumnDescription', 'name type_code display_size internal_size precision scale null_ok')
 """Named tuple for representing results from :attr:`Cursor.description`."""
 
 
--- a/python-phoenixdb/phoenixdb/tests/dbapi20.py
+++ b/python-phoenixdb/phoenixdb/tests/dbapi20.py
@@ -74,7 +74,7 @@ class DatabaseAPI20Test(unittest.TestCase):
     insert = 'insert'
 
     lowerfunc = 'lower' # Name of stored procedure to convert string->lowercase
-        
+
     # Some drivers may need to override these helpers, for example adding
     # a 'commit' after the execute.
     def executeDDL1(self,cursor):
@@ -208,7 +208,7 @@ class DatabaseAPI20Test(unittest.TestCase):
                 con.rollback()
             except self.driver.NotSupportedError:
                 pass
-    
+
     def test_cursor(self):
         con = self._connect()
         try:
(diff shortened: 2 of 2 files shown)


python-phoenixdb/phoenixdb/cursor.py
```diff
@@ -115,6 +115,7 @@ def description(self):
         description = []
         for column in self._signature.columns:
             description.append(ColumnDescription(
+                column.label,
```
stoty, line 118: Replace column_name, instead of  adding a new element.

python-phoenixdb/gen-protobuf.sh
```diff
@@ -16,7 +16,7 @@
 # limitations under the License.
 
 set -x
-AVATICA_VER=1.21.0
+AVATICA_VER=1.23.0
```
stoty, line 19: no need to rebuild protobuf, there have been no changes.

python-phoenixdb/.gitignore
```diff
@@ -8,6 +8,5 @@
 .tox
 dev-support/artifacts
 dev-support/work
-phoenixdb/.eggs
-phoenixdb/build
-phoenixdb/e
+.eggs
```
stoty, line 11: keep the old entries instead of replacing, some build scenarios may generate files there

stoty (changes requested):

stoty:
Pleas also add a test case for this (like the example mentioned in the ticket)

stoty:
In the test also check that we use the label as the key in  DictCursor as well.

python-phoenixdb/phoenixdb/cursor.py
```diff
@@ -27,7 +27,7 @@
 # TODO see note in Cursor.rowcount()
 MAX_INT = 2 ** 64 - 1
 
-ColumnDescription = collections.namedtuple('ColumnDescription', 'name type_code display_size internal_size precision scale null_ok')
+ColumnDescription = collections.namedtuple('ColumnDescription', 'label type_code display_size internal_size precision scale null_ok')
```
stoty, line 30: Nit: While the names are not part of the PEP249 API, we use the names defined in the standard as keys.
Please change this back to name.
(PEP249 has no concept of name vs label, the fact that JDBC driver calls the visible name "label" does not need to be visible here.)

python-phoenixdb/phoenixdb/tests/dbapi20.py
```diff
@@ -832,3 +832,18 @@ def test_ROWID(self):
...
+    def test_alias(self):
+        con = self._connect()
+        try:
+            cur = con.cursor()
+            self.executeDDL2(cur)
+            cur.execute("CREATE TABLE IF NOT EXISTS STOCK_SYMBOL (SYMBOL VARCHAR NOT NULL PRIMARY KEY, COMPANY VARCHAR)")
+            cur.execute("UPSERT INTO STOCK_SYMBOL VALUES ('CRM','SALESFORCE')")
+            cur.execute("UPSERT INTO STOCK_SYMBOL VALUES ('AAPL','APPLE Inc')")
+            cur.execute("SELECT symbol as symb, company as comp FROM STOCK_SYMBOL")
+            self.assertEqual(cur.description[0][0], 'SYMB')
```
stoty, line 846: Confirm that you can get the results via cur["SYMB"] and cur["COMP"]

python-phoenixdb/phoenixdb/tests/dbapi20.py
```diff
@@ -832,3 +832,18 @@ def test_ROWID(self):
         self.assertTrue(hasattr(self.driver,'ROWID'),
             'module.ROWID must be defined.'
             )
+
+    # https://issues.apache.org/jira/browse/PHOENIX-6917
+    def test_alias(self):
+        con = self._connect()
+        try:
+            cur = con.cursor()
```
stoty, line 840: use cursor_factory=phoenixdb.cursor.DictCursor

python-phoenixdb/phoenixdb/tests/dbapi20.py
```diff
@@ -832,3 +832,18 @@ def test_ROWID(self):
         self.assertTrue(hasattr(self.driver,'ROWID'),
             'module.ROWID must be defined.'
             )
+
+    # https://issues.apache.org/jira/browse/PHOENIX-6917
```
stoty, line 836: This file was adopted from an external test suite, and by convention tests here use the global 'barfly' and 'booze' tables.

Please either keep that convention, or move this test to test_db.py, and use the DatabaseTestCase plumbing for opening connections and cleaning up tables.

Doing either ensures that there are no tables left dangling after the test suite has run.

python-phoenixdb/phoenixdb/tests/dbapi20.py
stoty: You will probably need to change filed DictCursor uses to make that work.

stoty (changes requested):

python-phoenixdb/phoenixdb/tests/dbapi20.py
stoty: Otherwise re-running the tests on the same instance will fail.

stoty:
I have run some local tests, and for me the test passes without your change (i.e. with column_name).
In my tests, I received "BEVERAGE" in both columns in the protobuf reply.
I tested with PQS head and Phoenix 5.2.0-SNAPSHOT.

Having said that, the behaviour that I am seeing sounds like a bug in either Avatica or Phoenix, and I agree that we should use label instead of column_name.

However, we also need to make the same change here:
 https://github.com/apache/phoenix-queryserver/blob/56344687383b6622eb953101ff982678f8583895/python-phoenixdb/phoenixdb/cursor.py#L389

stoty:
The latest version breaks the AqlAlchemy tests.

Some metadata operations do not return a label field for the column metadata.

columns {
  ordinal: 10
  nullable: 2
  column_name: "COLUMN_SIZE"
  type {
    id: 5
  }
}

We need to check if label is defined, and if not, then use column_name (in both places)

satyakommula96:
@stoty how to run whole test suite

stoty:
You need to use 

mvn clean verify -am -pl phoenix-queryserver-it -Dtest=foo \
-Dit.test=QueryServerBasicsIT#startLocalPQS \
-Ddo.not.randomize.pqs.port=true -Dstart.unsecure.pqs=true -Pshade-javax-servlet

to start the test PQS

stoty:
Committed with fixes.

PR #126, opened 2023-04-13
PHOENIX-6924 Bump Junit, Jetty and Gson to the latest versions in que…

stoty:
…ryserver

Jetty: 9.4.51.v20230217
Gson: 2.10.1
Junit: 4.13.1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,8 +79,8 @@
         <zookeeper.version>3.5.8</zookeeper.version>
         <curator.version>2.12.0</curator.version>
 
-        <gson.version>2.2.4</gson.version>
-        <jetty.version>9.4.46.v20220331</jetty.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -93,7 +93,7 @@
 
         <!-- Test Dependency versions -->
         <mockito-all.version>1.8.5</mockito-all.version>
-        <junit.version>4.12</junit.version>
+        <junit.version>4.13.1</junit.version>
```
richardantal, line 96: I think there is 4.13.2 version for junit, is there a reason why we don't use that?

stoty: I wasn't aware of it.
Thanks.

stoty: Update to 4.13.2

richardantal (approved):

PR #140, opened 2023-10-31
PHOENIX-7091 Cleanup and ban log4j and related jars from phoenix-quer…

NihalJain:
…y-server

- Add maven-enforcer-plugin and use it to ban commons-logging and other logging framework related jars
- Remove all traces of log4j:log4j, org.slf4j:slf4j-log4j12 and commons-logging:commons-logging from dependency list
- Replace log4j.properties with log4j2.properties (copied from hbase project)
- Fix issue where IT logs where not getting logged in surefire reports due to "Failed to load class "org.slf4j.impl.StaticLoggerBinder"
  - See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
- Fix log4j duplicate dependency warning in pom.xml


NihalJain:
My bad, didnot fail locally as had a local 5.1.3 build with log4j remediated. Dropped patched jars from local repo. Will update PR with fix.

NihalJain:
Code is ready for review. Please review. Also, please let me know if banning jars seems too agressive (FYI HBase does the same). 
CC: @stoty, @richardantal, @chrajeshbabu

pom.xml
```diff
@@ -370,6 +370,49 @@
...
+                        <execution>
+                            <id>banned-other-logging-framework</id>
+                            <goals>
+                                <goal>enforce</goal>
+                            </goals>
+                            <configuration>
+                                <rules>
+                                    <bannedDependencies>
+                                        <excludes>
+                                            <exclude>log4j:*</exclude>
```
stoty, line 403: Do we even need to allow log4j2 ?
Isn't all logging happening via slf4j ?

pom.xml
```diff
@@ -815,23 +973,19 @@
...
+                    </exclusion>
+                </exclusions>
             </dependency>
             <dependency>
                 <groupId>bouncycastle</groupId>
                 <artifactId>bcprov-jdk15</artifactId>
                 <version>140</version>
                 <scope>test</scope>
             </dependency>
-            <dependency>
```
stoty, line 825: So these were duplicated.

stoty (approved):
+1 LGTM.

pom.xml
NihalJain: We use log4j2. slf4j is just a facade/abstraction layer and doesnot perform any logging in itself. Log4j2 on other hand provides actual logging implementation. So we do need both innthe project.

pom.xml
NihalJain: right!

pom.xml
stoty: I thought that these rules only apply to the source code references, and not to the transitive dependencies.
But I have never tested.

NihalJain: Ah we could just restric imports via https://github.com/skuzzle/restrict-imports-enforcer-rule. But maven-enforcer-plugin will ban transitive dependecies as well.

stoty: Thanks for the explanation.

PR #148, opened 2024-03-02
PHOENIX-7246: Add support for mapping JDBC type code 0 to protobuf NULL type

satyakommula96:


satyakommula96:
please review this PR @stoty

python-phoenixdb/phoenixdb/types.py
```diff
@@ -298,6 +298,8 @@ def from_column(column):
 
     @staticmethod
     def _from_jdbc(jdbc_code):
+        if jdbc_code == 0:
```
stoty, line 301: This is a bit ugly.
Would it be possible to add this to the type maps instead ?

satyakommula96: addressed

python-phoenixdb/phoenixdb/types.py
```diff
@@ -222,9 +225,9 @@ def __cmp__(self, other):
     (2011, common_pb2.STRING),  # NCLOB
     (2009, common_pb2.STRING),  # SQLXML
     # Returned by Avatica for Arrays in EMPTY resultsets
-    (2000, common_pb2.BYTE_STRING)  # JAVA_OBJECT
+    (2000, common_pb2.BYTE_STRING),  # JAVA_OBJECT
     # These are defined by JDBC, but cannot be mapped
-    # NULL
+    (0, common_pb2.NULL) # NULL
```
stoty, line 230: The comment is obviously not true for this one.
Please move this to the start (in numerical order)

stoty: They are not numerically ordered.
Still, it should be grouped with the other standard Phoenix types.

stoty:
Committed with minor cleanups.

Thanks a lot for the patch:

Some nits for the future:
- We do not use colon after the JIRA ID in the commit message
- The commit message should match the JIRA description
- Run flake8 / pylint on the code, it is discussed in the README

PR #149, opened 2024-03-05
PHOENIX-7256 Set java.io.tmpdir to the maven build directory for tests in PQS

Aarchy:


stoty (approved):
+1 LGTM

I didn't know you could set system properties like this for suerfire.

PR #150, opened 2024-03-08
PHOENIX-7266 Fix SSLConfigDir in ITs

Aarchy:


phoenix-queryserver-it/src/it/java/org/apache/phoenix/end2end/TlsUtil.java
```diff
@@ -51,6 +51,10 @@ public class TlsUtil {
     protected static final String TARGET_DIR_NAME = System.getProperty("target.dir", "target");
     protected static final File TARGET_DIR =
             new File(System.getProperty("user.dir"), TARGET_DIR_NAME);
+    protected static final String TEST_CLASSES_DIR_NAME =
```
stoty, line 54: A few lines of comments on what is happening here would be useful.

Aarchy: I'm not sure about this, as these are just directory names, there isn't really a complex logic behind them. There are very similar definitions in the preceding lines without comments as well.

richardantal (approved):
LGTM

PR #163, opened 2024-10-14
PHOENIX-7430 Remove commons-collections dependency in queryserver

NihalJain:
- Also remove imports and replace with equivalent code

### DIFF
--- a/phoenix-queryserver/src/main/java/org/apache/phoenix/util/InstanceResolver.java
+++ b/phoenix-queryserver/src/main/java/org/apache/phoenix/util/InstanceResolver.java
@@ -17,8 +17,7 @@
  */
 package org.apache.phoenix.util;
 
-import org.apache.commons.collections.IteratorUtils;
-
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceLoader;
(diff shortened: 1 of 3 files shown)


phoenix-queryserver/src/main/java/org/apache/phoenix/util/InstanceResolver.java
```diff
@@ -68,10 +67,13 @@ public static <T> T getSingleton(Class<T> clazz, T defaultInstance) {
     @SuppressWarnings("unchecked")
     public static <T> List get(Class<T> clazz, List<T> defaultInstances) {
         Iterator<T> iterator = ServiceLoader.load(clazz).iterator();
+        List<T> instances = new ArrayList<>();
```
NihalJain, line 70: There is no guava in PQS, so had to do this inplace using lambdas. Should we sync same logic in Phoenix? This class seems to be present at both places.

pom.xml
```diff
@@ -723,12 +722,6 @@
                   </exclusion>
                 </exclusions>
             </dependency>
```
NihalJain, line 724: We do not have restrict plugin in PQS. Should we add one here and sync all restricts or just add commons-collections restricts?

NihalJain:
Hi @stoty, could you please review?

pom.xml
stoty: It's up to you.
PQS has minimal code.
If you decide to add the restricions, then I think it's better to sync everything from the main repo.

stoty (approved):
+1 LGTM

phoenix-queryserver/src/main/java/org/apache/phoenix/util/InstanceResolver.java
NihalJain: sure makes sense

pom.xml
NihalJain: Yes make sense, let me raise another JIRA for same, this will keep changes in PR as per JIRA title

NihalJain: Raised PHOENIX-7438

PR #166, opened 2024-10-15
PHOENIX-7261 Align mockito version with Hadoop and HBase in QueryServer

NihalJain:


pom.xml
```diff
@@ -95,7 +95,7 @@
 
         <!-- Test Dependency versions -->
         <testng.version>6.10</testng.version>
-        <mockito-all.version>1.8.5</mockito-all.version>
+        <mockito-all.version>4.11.0</mockito-all.version>
```
stoty, line 98: We are not using mockito-all, better rename this to mockito.version

stoty (changes requested):
Let's clean up the property name, too.

pom.xml
NihalJain: sure @stoty let me update

NihalJain: done

stoty (approved):
+1 lgtm

PR #173, opened 2025-02-19
PHOENIX-7530 Support Python 3.13 in PhoenixDB

stoty:
also add workaround for pyenv virtualenv incompatibility


richardantal (approved):
LGTM

PR #176, opened 2025-04-01
PHOENIX-7523 Parse hbase-env.sh in PQS scripts, and add _OPTS environ…

stoty:
…ment variables for pqs and sqlline-thin


richardantal (approved):
LGTM

PR #179, opened 2025-05-07
PHOENIX-7600 - Replace commons-logging with slf4j

meszinorbi:


stoty (changes requested):
The new JAR also needs to be added the /lib directory in the assembly, and to the startup python scripts.

meszinorbi:
Did the requested changes.

stoty (changes requested):
The new jar is not added to the assembly.

stoty:
Please fix the orchestrator dependency check error.

probably copying this from phoenix-queryserver-it will fix it:

`      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-dependency-plugin</artifactId>
        <configuration>
          
          <ignoreNonCompile>true</ignoreNonCompile>
        </configuration>
      </plugin>`

meszinorbi:
Added the ignoreNonCompile true line to orchestrator, now it built locally with the  _mvn -B clean verify -DskipTests_ command, which was failing on github (without the -DskipTests).

PR #181, opened 2025-07-14
PHOENIX-7665 Make TLS protocols and ciphersuites configurable in PQS

stoty:


stoty:
This depdends on an Avatica feature that has not been released.
(Not even committed at the time of writing this)

PR #185, opened 2025-12-11
Support skip_autocommit_rollback option for sqlalchemy

artofhuman:
This option disables rollback attempts when the database is in autocommit mode.
It prevents errors that may occur when closing an invalid connection and also reduces the number of unnecessary queries.

doc: https://docs.sqlalchemy.org/en/21/core/pooling.html#disabling-reset-on-return-for-non-transactional-connections

sqlalchemy commit:
https://github.com/sqlalchemy/sqlalchemy/commit/eea260f5bed9f64a8ae9d993b4972b754524078a


artofhuman:
@stoty Could you take a look?

